PR #84, opened 2025-07-28
[Feat/LG-1963]: Checking multi branding support

shyam-1mg:


mayankmahavar1mg:
hey shyam, can you provide some kind of description and context regrading this feature.
What is the use case ?

PR #96, opened 2025-09-11
Universal App - Notifications (android)

mayansharma:


trinachoudhury1mg:
#review

src/native/androidProject/app/src/main/java/io/yourname/androidproject/MainActivity.kt
```diff
@@ -11,14 +12,23 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.cancelChildren
 import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
+import org.json.JSONObject
 
 class MainActivity : AppCompatActivity(), CoroutineScope by MainScope() {
 
+    companion object {
+        private var currentInstance: MainActivity? = null
```
vikramthakur3360, line 20: We should not keep static instances of Activity, this is a memory leak.

PR #100, opened 2025-09-22
New lifecycle, better error handling

mayansharma:
- Adds 2 new lifecycle methods - `onAppServerSideSuccess`, `onAppServerSideError`
- Error code fix in `renderMarkup`
- Same params in lifecycle methods

### DIFF
--- a/src/server/renderer/handler.js
+++ b/src/server/renderer/handler.js
@@ -25,6 +25,8 @@ import {
     onRouteMatch,
     onFetcherSuccess,
     onFetcherError,
+    onAppServerSideSuccess,
+    onAppServerSideError,
     onRenderError,
     onRequestError,
 } from "@catalyst/template/server/index.js"
@@ -182,7 +184,7 @@ const renderMarkUp = async (
     }
 
     try {
-        let status = matches.length && matches[0].match.path === "*" ? 404 : 200
+        let status = errorCode || (matches.length && matches[0].match.path === "*" ? 404 : 200)
         res.set({ "content-type": "text/html; charset=utf-8" })
         res.status(status)
 
@@ -202,7 +204,7 @@ const renderMarkUp = async (
                 onError(error) {
                     logger.error({ message: `\n Error while renderToPipeableStream : ${error.toString()}` })
                     // function defined by user which needs to run if rendering fails
-                    safeCall(onRenderError)
+                    safeCall(onRenderError, { req, res, store, error })
                     reject(error)
                 },
             })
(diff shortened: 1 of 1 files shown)


src/server/renderer/handler.js
```diff
@@ -182,7 +184,7 @@ const renderMarkUp = async (
     }
 
     try {
-        let status = matches.length && matches[0].match.path === "*" ? 404 : 200
+        let status = errorCode || (matches.length && matches[0].match.path === "*" ? 404 : 200)
```
trinachoudhury1mg: We'd have to revisit how we ship documents in cases of errors, right now it seems we are getting shell also shipped in cases of errors

src/server/renderer/handler.js
```diff
@@ -255,7 +257,7 @@ export default async function (req, res) {
         let allTags = []
 
         // function defined by user which needs to run after route is matched
-        safeCall(onRouteMatch, { req, res, matches })
+        safeCall(onRouteMatch, { req, res, matches, store })
```
trinachoudhury1mg: how to we know the route matched?

src/server/renderer/handler.js
```diff
@@ -280,24 +285,65 @@ export default async function (req, res) {
                     return Promise.resolve(res)
```
trinachoudhury1mg: little concerned with the placement of this check, as we are resolving promise here while we are taking out err in next line.

src/server/renderer/handler.js
```diff
@@ -280,24 +285,65 @@ export default async function (req, res) {
...
                 }
-
-                return new Promise((resolve, reject) => {
-                    renderMarkUp(null, req, res, allTags, fetcherData, store, matches, context, webExtractor)
-                        .then(resolve)
-                        .catch(reject)
-                })
             } catch (error) {
-                // TODO: serverDataFetcher never throws any error
+                // TODO: This catch block is kept for any unexpected errors from serverDataFetcher or getMetaData
```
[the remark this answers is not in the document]
mayansharma: This block doesn't run for errors thrown by `serverDataFetcher`, but is still required for `getMetaData`, no changes required as of now

src/server/renderer/handler.js
mayansharma: The `matches` array comes empty

src/server/renderer/handler.js
mayansharma: This is required for cases in which the developer redirects / ends response inside their `serverFetcher`

src/server/renderer/handler.js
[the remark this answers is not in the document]
trinachoudhury1mg: lets update the comment then @mayansharma

src/server/renderer/handler.js
mayansharma: We call this lifecycle irrespective of route matches, because our current flow proceeds independent of route matching

PR #135, opened 2025-11-17
Feature: Build Version Management for Android Builds

Uttkarsh-Srivastava:
### Overview
Added build version prompting for Android builds. User-provided build version is stored in `webview_config.properties` and accessible via the `deviceInfo` API through the native-web bridge.

#### 1. CLI Prompt
- User is prompted for build version when running `npm run buildApp:android`
- Accepts any string value (no format validation)
- Defaults to `0.0.1` if left empty
- Applied to both debug and release builds

#### 2. Build Pipeline Integration
- Build version passed through entire Android build pipeline
- Gradle receives version via `-PbuildVersion` parameter
- Written to `webview_config.properties` during build

#### 3. Device Info API Enhancement
- Android `getDeviceInfo()` now returns `buildVersion` field
- Read from `webview_config.properties` at runtime
- Updated WebBridge documentation with new field
- iOS not implemented (Android-only feature)

### Files Modified
- `src/native/utils.js` - Added `promptBuildVersion()` utility
- `src/native/buildAppAndroid.js` - Integrated version prompting
- `src/native/androidProject/app/build.gradle.kts` - Accept and store buildVersion
- `src/native/androidProject/app/src/main/java/.../DeviceInfoUtils.kt` - Include buildVersion in response
- `src/native/androidProject/app/src/main/java/.../NativeBridge.kt` - Pass properties to DeviceInfoUtils
- `src/native/bridge/WebBridge.js` - Updated documentation

### Usage Flow
1. Developer runs `npm run buildApp:android`
2. CLI prompts: `"Enter build version (default: 0.0.1): "`
3. User enters version or presses Enter for default
4. Version passed to build pipeline
5. Gradle writes to `webview_config.properties`
6. Android app reads and exposes via `getDeviceInfo()` API

### Example DeviceInfo Response
{
  "model": "Pixel 6",
  "manufacturer": "Google",
  "platform": "android",
  "screenWidth": 1080,
  "screenHeight": 2400,
  "screenDensity": 2.75,
  "buildVersion": "1.0.0"
}### Notes
- No version format validation - accepts any string
- No version history tracking
- Empty input automatically uses default value "0.0.1"

### DIFF
--- a/src/native/utils.js
+++ b/src/native/utils.js
@@ -186,79 +186,13 @@ async function validateAndCompleteConfig(platform, configPath) {
     }
 }
 
-// Version history management functions
-function loadBuildVersionHistory(buildVersionsFile) {
-    try {
-        if (fs.existsSync(buildVersionsFile)) {
-            const content = fs.readFileSync(buildVersionsFile, "utf8")
-            return JSON.parse(content)
-        }
-    } catch (error) {
-        console.warn(`Warning: Could not load version history: ${error.message}`)
-    }
-    // Return default structure
-    return {
-        android: {
-            versions: [],
-            lastUsed: null,
-        },
-    }
-}
+// Build version prompt
+async function promptBuildVersion(logger = console) {
+    const version = await promptUser("\nEnter build version (default: 0.0.1): ")
+    const finalVersion = version || "0.0.1"
 
-function saveBuildVersionHistory(buildVersionsFile, history) {
-    try {
-        fs.writeFileSync(buildVersionsFile, JSON.stringify(history, null, 4), "utf8")
-    } catch (error) {
-        throw new Error(`Failed to save version history: ${error.message}`)
-    }
-}
-
-function isVersionUsed(version, history) {
-    return history.android.versions.includes(version)
-}
-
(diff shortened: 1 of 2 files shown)


src/native/utils.js
```diff
@@ -186,11 +186,91 @@ async function validateAndCompleteConfig(platform, configPath) {
...
+    const history = loadBuildVersionHistory(buildVersionsFile)
+
+    let isValid = false
+    let version = ""
+
+    while (!isValid) {
+        version = await promptUser("\nEnter build version (format: X.Y.Z, e.g., 1.0.0): ")
+
+        // Validate format
+        if (!versionRegex.test(version)) {
```
mayankmahavar1mg: are we sure about this logic,

1.0.0-abc not a valid build Version ??

if not this error must be coming from buildCommand. we should use that,
not built custom logic

there is possibilty here what we validate can be invalid for gradle and vice versa

Uttkarsh-Srivastava:
> Are there plans to add buildversion enhancement to the `getdeviceinfo` api on iOS?

@sauravvarma There was a requirement to attach a version to every build generated by the developer. This PR just does that. The buildVersion enhancement we discussed would be picked later.

PR #141, opened 2025-11-21
Universal App - Combined branch (Notifications + Warnings + XCConfig)

mayansharma:


mayankmahavar1mg:
1. The topic subscription callback mismatch - This is a clear bug. Should we fix it to use TOPIC_SUBSCRIPTION_RESULT event?
  2. Push registration timeout - Android has no timeout. What if FCM initialization hangs? iOS has 30s timeout. Should Android match?
  3. Error handling - The React hook doesn't have dedicated error callbacks registered. Errors go to base.handleNativeError() but some Android methods send errors via wrong events.
  4. The /notification route - Does the Catalyst app in the consuming project handle this route? Or is there middleware in catalyst-core that parses ?action=...&data=... - what kind of request is going ??
  5. (cancelAll) - This is definitely a bug. When user taps one notification, it dismisses ALL app notifications. Should track this for fix.
  6. android has duplicacy vs ios , we can remove this duplicacy
  ---
  📊 Comparison Table

  | Aspect             | Android (Current)      | iOS (Current)             | SIngle repo (Android) |
  |--------------------|------------------------|---------------------------|-------------------------|
  | Code Duplication   | 849 lines duplicated   | 0 lines duplicated        | 0 lines duplicated      |
  | Maintenance        | Fix bugs in 2 places   | Fix once                  | Fix once                |
  | Runtime Check      | ❌ Build-time only      | ✅ Checks plist at runtime | ✅ Can check at runtime  |
  | Firebase Detection | ❌ Build decision       | ✅ Runtime detection       | ✅ Class.forName check   |
  | Graceful Fallback  | ⚠️ Via separate build  | ✅ Automatic               | ✅ Automatic             |
  | Module Separation  | ❌ Same source tree     | ✅ Separate modules        | ⚠️ Same package         |
  | APK Size           | Same (gradle excludes) | Same (SPM excludes)       | Same                    |

  ---

PR #143, opened 2025-11-25
Otel: add support for grpc based trace endpoints

vishalpolley-1mg:
Otel: add support for grpc based trace endpoints

### DIFF
--- a/src/otel.js
+++ b/src/otel.js
@@ -30,7 +30,7 @@ function init(config = {}) {
         metricHeaders = {},
         batchProcessorConfig = {},
         exportIntervalMillis = 10000,
-        instrumentations = [],
+        instrumentations,
         samplingRate = 1.0,
         grpcCredentials,
     } = config


src/otel.js
```diff
@@ -18,49 +22,63 @@ function init(config = {}) {
...
-                scheduledDelayMillis: exportIntervalMillis,
-                maxQueueSize: 100,
-                maxExportBatchSize: 10,
-            }),
-            metricReader: new PeriodicExportingMetricReader({
-                exporter: otlpMetricExporter,
-                exportIntervalMillis,
-            }),
+            spanProcessor: new BatchSpanProcessor(otlpTraceExporter, batchProcessorConfig),
             instrumentations: instrumentations ?? [getNodeAutoInstrumentations()],
```
mayansharma: Would this work if no `instrumentations` is passed? We have a `[]` fallback when extracting this from the config

vishalpolley-1mg: removing the default parameter as []

src/otel.js
```diff
@@ -82,6 +100,64 @@ function init(config = {}) {
...
+ * @param {string} protocol - "grpc" or "http"
+ * @param {string} url - Exporter endpoint URL
+ * @param {object} headers - Headers to include in requests
+ * @param {Function} [grpcCredentials] - gRPC Credentials (optional)
+ * @returns {OTLPTraceExporter|OTLPTraceExporterHTTP} Configured trace exporter
+ */
+function createTraceExporter(protocol, url, headers = {}, grpcCredentials) {
+    if (protocol.toLowerCase() === "http") {
+        logger.info(`📡 Creating HTTP trace exporter for URL: ${url}`)
+        return new OTLPTraceExporterHTTP({
```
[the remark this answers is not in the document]
vishalpolley-1mg: yes, the user can pass manual batchProcessorConfig, also by default BatchSpanProcessor has these configurations defined, so user can override manually if needed.

Screenshot 2025-11-20 at 5 23 41 PM

PR #216, opened 2026-03-31
Add otel tracing around handler

Uttkarsh-Srivastava:


Mrsandeep27 (commented):
Clean approach to threading tracing through the hook lifecycle. A few notes:

**1. `tracedResWriteFirstFoldCss` / `tracedResWriteFirstFoldJS`** — both are just thin wrappers around `res.write(chunk)` with the same signature. Consider a single `tracedResWrite` factory (or pass the span name as an argument) so you don't duplicate the closure for each chunk type. Current setup will also double-count if you ever add a third chunk.

**2. `traceHandlerHook` gates on `typeof fn === "function"`.** If a template exports `onRouteMatch` as `undefined` (common when a project only opts into some hooks), the fallthrough returns `fn` unchanged. Good. But you then pass that `undefined` into `safeCall(onRouteMatch, ...)` — confirm `safeCall` handles `undefined` fn without emitting a misleading "Invalid lifecycle method" line for hooks that are legitimately optional.

**3. `_handler` is now internal-only** but it's still `async`. `withObservability` must preserve the promise chain for the rejection path (`throw (error) => Promise.reject(error)` at the bottom). Worth confirming the wrapper awaits or forwards the rejection — otherwise you'll see tracing spans without error status even when the real handler rejected.

**4. Unrelated changes** — `"Google Agent":"Google-Agent"` addition in `userAgentUtil.js` and the indentation touch on line 239 (`     const isBot` → `    const isBot`) look like drive-by edits. Splitting them into their own commits would make this PR cleaner and easier to revert if tracing needs a rollback later.

**5. Nit:** Span names like `res.write.firstFoldCss` use dot notation, but the existing spans in this file use camelCase (`getMetaData`, `renderMarkUp`, `onRouteMatch`). Picking one style keeps trace dashboards scannable.

Direction looks right; mostly cleanup suggestions.

PR #217, opened 2026-04-01
Fixed eslint violations across MCP v2 tools and native hook bridge

mayansharma:
Clean up ESLint violations across the MCP v2 stack and native bridge hook utilities. This removes unused variables, replaces empty catch blocks with explicit no-op handling, lifts an inner helper declaration in mcp.js, and fixes the missing useCallback dependency in useBaseHook so the repo passes eslint . again.


trinachoudhury1mg (approved):
LGTM

PR #229, opened 2026-04-30
Added bot detection to the request middleware flow

D-Agam:
checking isBot in req for automatic detection instead of recalculating bot details from the user agent.


D-Agam-tech:
Any future bot list changes can be made in the 1mg_web repo without duplicating them in catalyst-core.
Adding isBot to the request in addDeviceDetails in 1mg_web so Catalyst renderer can reuse the bot detection result.

trinachoudhury1mg:
@Uttkarsh-Srivastava - Merging this but we need to cleanup bot logic once and for all to cater access to users to provide customization be it list level or function level to drive bot logic

PR #252, opened 2026-05-22
fix(serve): hard-fail when build is stale vs current config

### LINKED ISSUE #251: catalyst serve silently serves stale build when config.json changes between build and serve
## Summary

`catalyst serve` happily serves a build whose embedded asset host no longer matches `config/config.json`. The boot log prints the new host, while the HTML it serves still references chunks at the old host. The result on a device is "page loads, no JS/CSS" with no obvious signal from the serve process.

This is easy to hit when developing on a laptop whose local IP changes (Wi-Fi switch, tethering on/off, etc.) and you update `PUBLIC_STATIC_ASSET_URL` but forget to rebuild before `npm run serve`.

## Repro

1. Set `PUBLIC_STATIC_ASSET_URL` to `http://A:3005` in `config/config.json`.
2. `npm run build`.
3. Edit config so `PUBLIC_STATIC_ASSET_URL` is `http://B:3005`.
4. `npm run serve`.
5. Server logs `Local: http://B:3005`.
6. `curl http://B:3005/` — the HTML references `http://A:3005/assets/...` for `runtime.*.js` and all chunks.

## Why it happens

In `packages/catalyst-core/src/webpack/base.babel.js`, webpack's `publicPath` and `DefinePlugin` bake `PUBLIC_STATIC_ASSET_URL` and `PUBLIC_STATIC_ASSET_PATH` into the client bundle at build time. `serve.js` re-reads `config.json` at startup and uses it for the boot log + `NODE_SERVER_HOSTNAME`/`PORT`, but never checks whether the values that were embedded in the build still match the current config. So the two diverge silently.

The misleading log makes the misconfig hard to spot — everything looks healthy on the serve side; only an inspection of the served HTML reveals the bug.

## Proposed fix

Have `build` persist the values that webpack embeds, and have `serve` refuse to start when the current config diverges from the embedded values.

PR: https://github.com/tata1mg/catalyst-core/pull/252

## Notes / open questions

- Tracked keys are scoped to the values that actually get embedded into client assets (`PUBLIC_STATIC_ASSET_URL`, `PUBLIC_STATIC_ASSET_PATH`). `CLIENT_ENV_VARIABLES` values are also embedded via DefinePlugin and arguably belong in the check — open to extending the tracked set if maintainers prefer.
- Hard-fail vs warn-and-continue: the PR hard-fails, since silently serving a broken build is exactly the bug. Happy to make it a `--allow-stale-build` opt-out if hard-fail is too aggressive.

sauravvarma:
## Summary

- Tracks `PUBLIC_STATIC_ASSET_URL` and `PUBLIC_STATIC_ASSET_PATH` in `BUILD_OUTPUT_PATH/build-meta.json` at build time (the values webpack embeds into the client bundle via `publicPath` + `DefinePlugin`).
- On `catalyst serve` startup, compares the embedded values against the current `config/config.json` and exits with a clear diff message on mismatch.
- Missing `build-meta.json` (older builds) emits a warning and proceeds — no break for existing deployments.

Closes #251.

## Why

`catalyst serve` currently re-reads `config.json` for the boot log and `NODE_SERVER_HOSTNAME`/`PORT`, but the asset host that ends up in the served HTML was frozen at build time. When config changes between build and serve (common when local IP changes), serve logs `Local: http://<new-host>` while the HTML still references chunks at the old host. On a device this looks like "page loads, no JS/CSS" with no signal from the serve process.

## Behavior

Before (config changed from `A` to `B` after the build, then `npm run serve`):
```
Local: http://B:3005
```
…and the HTML embeds `http://A:3005/assets/runtime.*.js`. Silent stale serve.

After:
```
[serve] Build is stale relative to config/config.json.

The following values were baked into the build but differ from the current config:

  PUBLIC_STATIC_ASSET_URL
    built with: http://A:3005
    config now: http://B:3005

These values are embedded into client assets at build time, so serving this build
would reference assets at the old host. Run npm run build to rebuild.
```
Exit code 1.

## Test plan

- [ ] `npm run build` in a fixture app writes `build/build-meta.json` with current `PUBLIC_STATIC_ASSET_URL` / `PUBLIC_STATIC_ASSET_PATH`.
- [ ] `npm run serve` with matching config starts normally.
- [ ] Mutate `PUBLIC_STATIC_ASSET_URL` in `config/config.json` and re-run `npm run serve` — exits non-zero with the diff message above, no server process spawned.
- [ ] Delete `build/build-meta.json` and re-run `npm run serve` — emits the "skipping check" warning and proceeds (backwards-compat with builds from prior versions).

## Notes / open questions

- Hard-fail is intentional: the bug is precisely that the broken state was silent. Happy to convert to warn-and-continue or gate behind a `--allow-stale-build` flag if maintainers prefer.
- Currently only the two `PUBLIC_STATIC_ASSET_*` keys are tracked. `CLIENT_ENV_VARIABLES` values are also embedded via `DefinePlugin`; extending the check to those is a small follow-up if desired.

### DIFF
--- a/packages/catalyst-core/src/scripts/build.js
+++ b/packages/catalyst-core/src/scripts/build.js
@@ -1,8 +1,27 @@
+const fs = require("fs")
 const path = require("path")
 const { green, cyan, yellow } = require("picocolors")
 const { name } = require(`${process.cwd()}/package.json`)
-const { BUILD_OUTPUT_PATH } = require(`${process.cwd()}/config/config.json`)
+const appConfig = require(`${process.cwd()}/config/config.json`)
+const { BUILD_OUTPUT_PATH, PUBLIC_STATIC_ASSET_URL, PUBLIC_STATIC_ASSET_PATH } = appConfig
 const { arrayToObject, printBundleInformation, runBuildCommands } = require("./scriptUtils.js")
+
+const BUILD_META_FILENAME = "build-meta.json"
+
+/**
+ * @description - Persists the config values that webpack bakes into the bundle so
+ * that `serve` can detect when the build is stale relative to the current config.
+ * Only keys whose values get embedded into client assets at build time are tracked here.
+ */
+function writeBuildMeta() {
+    const meta = {
+        PUBLIC_STATIC_ASSET_URL,
+        PUBLIC_STATIC_ASSET_PATH,
+        builtAt: new Date().toISOString(),
+    }
+    const metaPath = path.join(process.cwd(), BUILD_OUTPUT_PATH, BUILD_META_FILENAME)
+    fs.writeFileSync(metaPath, JSON.stringify(meta, null, 2))
+}
 
 /**
  * @description - creates a production build of the application.
@@ -36,6 +55,8 @@ function build() {
             ...argumentsObject,
         },
     })
+
+    writeBuildMeta()
 
     console.log(green("Compiled successfully."))
     console.log("\nFile sizes after gzip:\n")
--- a/packages/catalyst-core/src/scripts/serve.js
+++ b/packages/catalyst-core/src/scripts/serve.js
@@ -1,8 +1,64 @@
+const fs = require("fs")
 const path = require("path")
 const { spawnSync } = require("child_process")
+const { red, yellow, cyan } = require("picocolors")
 const { arrayToObject } = require("./scriptUtils")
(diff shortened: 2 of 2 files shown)


Uttkarsh-Srivastava:
LGTM 👍🏽. This can be merged. This has conflicts with main though.

PR #254, opened 2026-05-25
feat(native/bridge): web fallback hooks + useNativeTransition (Android & iOS)

mayankmahavar1mg:
## Summary

Two major additions on top of `feature/video_hook_android_ios`:

1. **All 9 native hooks refactored into per-file structure with web fallbacks**
2. **`useNativeTransition` — snapshot overlay page transitions (Android + JS)**

---

## 1. Hooks refactor + web fallbacks

All hooks moved from the monolithic `hooks.js` into `src/native/bridge/hooks/`
with individual files per hook.

**Hooks with web fallbacks added:**
- `useVideoStream` — web: `getUserMedia` stream
- `useCamera` — web: `<input type=file capture=environment>` (no getUserMedia — iOS WebView safe)
- `useFilePicker` — web: `<input type=file>`
- `useDataProtection` — web: localStorage/sessionStorage
- `useHapticFeedback` — web: no-op (API unavailable in browser)
- `useGoogleSignIn` — web: Google Identity Services SDK
- `useIntent` — web: `window.open`
- `useSafeArea` — web: CSS env() vars
- `useNetworkStatus` — web: `navigator.onLine` + Network Information API

**`useNotification` — web fallback intentionally skipped** (Browser Push requires
Service Worker; unreliable in iOS WebView).

**API surface added to every hook:**
```js
const { ..., webFallbackActive, webFallbackDisabled, setWebFallback } = useXxx()
```
`webFallback` prop is synchronous source of truth; overrides internal state.

**Bug fixes included:**
- `useBaseHook`: `isNative` detection fixed to use `nativeBridge.isAvailable()`
  instead of `window.WebBridge` check (was always `false` on real Android devices)
- Rules of Hooks violations fixed in `useDataProtection`, `useFilePicker`, `useCamera`
- `isNative` / `isWeb` added to thin-state hooks (`useSafeArea`, `useNetworkStatus`, `useDeviceInfo`)
- `hooks.js` retained as re-export shim for backwards compatibility

---

## 2. useNativeTransition

Snapshot overlay pattern for native-quality page transitions:

1. JS calls `navigate(to, opts)` → bridge `startTransition`
2. Native snapshots current WebView, places bitmap overlay over full screen
3. JS router navigates underneath (snapshot hides the content swap)
4. JS calls `commitTransition` → native animates overlay out
5. Per-call safety timeout on native auto-commits if JS never calls commit

Web fallback: CSS overlay + `requestAnimationFrame` double-frame commit.

**Android:** `TransitionManager.kt` — Canvas bitmap, `ImageView` on `decorView`,
`Handler` safety timer, slide/fade `ObjectAnimator`. Bridge wired in
`NativeBridge.kt` + `BridgeUtils.kt`.

**JS:** `useNativeTransition.js` hook, `NativeBridge.js` helpers, command
registrations in `NativeInterfaces.js`.

```js
navigate('/path', {
  type: 'slide',      // 'slide' | 'fade'
  direction: 'left',  // 'left' | 'right'
  duration: 300,
  timeout: 900,       // safety ms, default max(duration*3, 800)
})
```

**iOS implementation complete but not included in this PR** (pending iOS-only commit).

---

## Files changed

| Area      | Files                                                       |
| --------- | ----------------------------------------------------------- |
| Android   | `NativeBridge.kt`, `BridgeUtils.kt`, `TransitionManager.kt` |
| JS bridge | `NativeBridge.js`, `NativeInterfaces.js`, `useBaseHook.js`  |
| Hooks     | `hooks/` directory (11 files), `hooks.js` shim              |
| KB        | `knowledge-base.json` (114 entries)                         |
| ```       |                                                             |


packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+            isLoading: base.loading,
+            clearPhoto: base.clear,
+        }
+    }
+
+    useEffect(() => {
+        window.WebBridge.register(NATIVE_CALLBACKS.ON_CAMERA_CAPTURE, (data) => {
+            try {
+                const result = typeof data === "string" ? JSON.parse(data) : data
+                console.log("📷 Camera capture result:", result)
```
trinachoudhury1mg, line 110: debug log? (check logs across)

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+            photo: null,
+            takePhoto: () => {},
+            isLoading: false,
+            clearPhoto: () => {},
+        }
+    }
+
+    // Web fallback callback — always declared (Rules of Hooks)
+    const webTakePhoto = useCallback(() => {
+        const input = document.createElement("input")
```
trinachoudhury1mg: when are we cleaning up the input element from DOM?

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+}
+
+export const useCameraPermission = ({ webFallback } = {}) => {
+    const isWeb = typeof window === "undefined" ? true : !nativeBridge.isAvailable()
+
+    // All hooks unconditionally at top level (Rules of Hooks)
+    const [webFallbackState, setWebFallback] = useState(true)
+    const webFallbackResolved = webFallback !== undefined ? webFallback !== false : webFallbackState
+    const webFallbackActive = isWeb && webFallbackResolved
+    const webFallbackDisabled = isWeb && !webFallbackResolved
```
trinachoudhury1mg, line 238: why 4 variables and role of each?

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+            permission: null,
+            photo: base.data,
+            takePhoto: webTakePhoto,
+            isLoading: base.loading,
+            clearPhoto: base.clear,
+        }
+    }
+
+    useEffect(() => {
+        window.WebBridge.register(NATIVE_CALLBACKS.ON_CAMERA_CAPTURE, (data) => {
```
trinachoudhury1mg, line 107: do we register callbacks even if its web and not app webview?

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+    return {
+        data: base.data,
+        loading: base.loading,
+        progress: base.progress,
+        error: base.error,
+        isWeb: base.isWeb,
+        isNative: base.isNative,
+        webFallbackActive: false,
+        webFallbackDisabled: false,
+        setWebFallback: base.setWebFallback,
```
trinachoudhury1mg, line 184: webFallbackActive: false,
        webFallbackDisabled: false,
        setWebFallback: base.setWebFallback,

why not just 1 flag for webFallbackEnabled - which determines whether on web , webfallback needs to be used directly

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        requestPermission()
+
+        return () => {
+            if (window.WebBridge && nativeBridge.isAvailable()) {
+                window.WebBridge.unregister(NATIVE_CALLBACKS.CAMERA_PERMISSION_STATUS)
+            }
+        }
+    }, [webFallbackDisabled])
```
trinachoudhury1mg, line 321: webFallbackDisabled - not a state variable to monitor useEffect upon

packages/catalyst-core/src/native/bridge/hooks/useDataProtection.js
```diff
@@ -0,0 +1,218 @@
...
+        if (window.indexedDB?.databases) {
+            const dbs = await window.indexedDB.databases().catch(() => [])
+            await Promise.all(dbs.map((db) => new Promise((res) => {
+                const req = window.indexedDB.deleteDatabase(db.name)
+                req.onsuccess = req.onerror = res
+            })))
+        }
+    }, [])
+
+    if (typeof window === "undefined") {
```
trinachoudhury1mg, line 60: repetitive returnables across, see if a function helps here with overrides as arg

packages/catalyst-core/src/native/bridge/hooks/useDataProtection.js
```diff
@@ -0,0 +1,218 @@
...
+            overlay.style.display = document.hidden ? "block" : "none"
+        }
+        document.addEventListener("visibilitychange", onVisibilityChange)
+
+        return () => {
+            document.removeEventListener("visibilitychange", onVisibilityChange)
+            overlay.remove()
+            secureOverlayRef.current = null
+        }
+    }, [base.webFallbackActive, screenSecureWeb])
```
trinachoudhury1mg, line 35: does base reference change during re-renders ?

packages/catalyst-core/src/native/bridge/hooks/useDataProtection.js
```diff
@@ -0,0 +1,218 @@
...
+                    base.handleNativeError(result.error)
+                    return
+                }
+                setScreenSecureState(Boolean(result?.secure))
+            } catch (error) {
+                base.handleNativeError(error)
+            }
+        }
+
+        const cleanup = registerNativeHandlers([
```
trinachoudhury1mg, line 178: i have a question, as a dev who wants to manage both for native app and web users he'd have to manage callbacks on their own by adding checks instead of having one set of events which internally identifies at our end whether its called for native or web. 
_General comment for all hooks where this is applicable._

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
mayankmahavar1mg: These logs are intentionally scoped to the native path — they're inside the if (!window.WebBridge) return guard (line 105), so they only fire when WebBridge is present (i.e., native app webview). They'll never show on plain web. Happy to wrap them in isDevelopment() if we want to mute them in prod native builds, but they serve as confirmation that the bridge callback is firing correctly.

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
mayankmahavar1mg: The useEffect (line 104) has an early return if (!window.WebBridge) return, so window.WebBridge.register(...) is only ever called when the native bridge is present. Pure web environments exit immediately and no callbacks are registered.

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
mayankmahavar1mg: These aren't defined in useCamera.js — they live in useBaseHook.js (lines 29–31). The logic is:
- webFallbackResolved — prop wins over state (synchronous, no async gap)
- webFallbackActive — is web AND fallback enabled
- webFallbackDisabled — is web AND fallback explicitly disabled

useCamera just passes { hasWebFallback: true, webFallback } to useBaseHook and consumes these computed flags.

packages/catalyst-core/src/native/bridge/hooks/useCamera.js
mayankmahavar1mg: On consolidation: webFallback prop being undefined (not passed) vs false (explicitly disabled) are semantically different — "use hook default" vs "explicitly disable fallback". A single flag loses that distinction. webFallbackResolved is the merge point, and webFallbackActive/webFallbackDisabled are the two exit conditions for early returns.

 On the dep: webFallbackDisabled is derived from webFallbackState (React state via useState) + isWeb (stable). Its value changes only when webFallbackState changes, so the effect re-runs correctly. It's a derived value but its source is state.

packages/catalyst-core/src/native/bridge/hooks/useDataProtection.js
mayankmahavar1mg: On consolidation: webFallback prop being undefined (not passed) vs false (explicitly disabled) are semantically different — "use hook default" vs "explicitly disable fallback". A single flag loses that distinction. webFallbackResolved is the merge point, and webFallbackActive/webFallbackDisabled are the two exit conditions for early returns.

 On the dep: webFallbackDisabled is derived from webFallbackState (React state via useState) + isWeb (stable). Its value changes only when webFallbackState changes, so the effect re-runs correctly. It's a derived value but its source is state.

packages/catalyst-core/src/native/bridge/hooks/useDataProtection.js
mayankmahavar1mg: This is already handled internally — the hook exposes a single execute (aliased as takePhoto for camera) that routes to the correct implementation based on environment. On native it calls nativeBridge.camera.open() and listens via WebBridge callbacks; on web it triggers the <input type="file"> fallback. The consumer just calls execute() or takePhoto() without any isNative/isWeb checks on their end.

 The isNative / isWeb flags are exposed for informational use (e.g. rendering different UI), not for gating the operation itself. Same pattern applies across all hooks that have a web fallback — useFilePicker, etc. all expose a single execute.

mayankmahavar1mg:
already merged

PR #261, opened 2026-06-02
feat: add parallel fetches and router level caching

gurditt-khurana:
Added parallel fetches for route bundle and route fetchers, also added support for config-driven route fetcher data caching


gurditt-khurana:
#review

PR #272, opened 2026-06-12
feat: add GitHub feedback integration to Catalyst MCP v2

pallavjain1mg:
### 📌 Overview
This PR introduces GitHub integration to the Catalyst MCP v2 server. It enables developers to submit bugs, feature requests, and general feedback directly to the `catalyst-core` repository from their AI assistant (Claude Desktop / DeputyDev). 

To ensure safety and quality, the implementation features a preview step and duplicate detection before anything is published.

---

#### 🛠️ New Tools Exposed:
1. **`preview_github_feedback`**: Drafts the issue/discussion content, automatically appends project metadata, searches GitHub for duplicates, and returns a preview for developer approval. *Must be called before publishing.*
2. **`create_github_issue`**: Creates a GitHub issue via the REST API (used for bugs and feature requests).
3. **`create_github_discussion`**: Creates a GitHub discussion via the GraphQL API (used for Q&A, proposals, and general discussion).

#### 🛡️ Key Features:
* **Zero External Dependencies**: Implemented using only native Node.js `https` and `child_process` modules (no package bloat).
* **Flexible Authentication**: Prioritizes local GitHub CLI (`gh auth token`) to prevent storing secrets, with a fallback to a `GITHUB_TOKEN` environment variable.
* **Smart Duplicate Detection**: Scans existing issues (REST) and discussions (GraphQL) before posting to prevent duplicate threads.
* **Automatic Metadata Triage**: Automatically appends the active project name, path, and `catalyst-core` package version to the footer of every submission.

---

### 🔑 Authentication Setup

The MCP server supports two authentication methods (evaluated in order):

1. **GitHub CLI (Recommended - Zero Config)**
   * Uses your active local session. Run `gh auth login` in your terminal.
   * *Required scopes:* `repo`, `write:discussion`.
2. **Environment Variable**
   * Pass `GITHUB_TOKEN` (Personal Access Token) in your MCP client environment configuration.
   * *Required scopes:* `repo`, `write:discussion`.

---

### 📂 Files Changed

* **[`mcp.js`](file:///Users/pallavjain/Documents/1mg_projects/catalyst-core/packages/catalyst-core/mcp_v2/mcp.js)**: Registered the new GitHub tools and wired them into the core MCP tool router with intent safeguards.
* **[`tools/github.js`](file:///Users/pallavjain/Documents/1mg_projects/catalyst-core/packages/catalyst-core/mcp_v2/tools/github.js)**: *(New)* Encapsulates the entire GitHub API integration (REST + GraphQL) and credentials handling.

### DIFF
--- a/packages/catalyst-core/mcp_v2/tools/github.js
+++ b/packages/catalyst-core/mcp_v2/tools/github.js
@@ -4,11 +4,10 @@
  * tools/github.js - Catalyst MCP v2
  *
  * GitHub issue workflow for AI agents:
- *   1. start_github_issue_flow      - tells the LLM what details to collect
- *   2. gather_github_issue_context  - reads project/framework context for triage
- *   3. preview_github_feedback      - formats the issue and checks duplicate candidates
- *   4. create_github_issue          - publishes after duplicate review, with markdown fallback
- *   5. generate_issue_markdown      - manual fallback when auth/API is unavailable
+ *   create_github_issue({ dry_run: true })  - first asks the developer to select labels
+ *   create_github_issue({ dry_run: true, labels }) - gathers context, templates, and duplicate candidates
+ *   create_github_issue({ dry_run: false }) - publishes after explicit developer approval
+ *   markdown fallback is generated internally when auth/API publishing fails
  *
  * Authentication:
  *   1. Direct GitHub API token from GITHUB_TOKEN, GITHUB_PAT, or CATALYST_GITHUB_TOKEN
@@ -228,6 +227,9 @@ function githubRequest(method, urlPath, token, body) {
         })
 
         req.on("error", reject)
+        req.setTimeout(30000, () => {
+            req.destroy(new Error("GitHub API request timed out after 30 seconds"))
+        })
         if (payload) req.write(payload)
         req.end()
     })
@@ -282,6 +284,38 @@ function resolveIssueLabels(args, body) {
     if (labels.length > 0) return labels
 
     return [resolveIssueTemplate(args, body)]
+}
+
+function hasLabelsInput(args) {
+    return Object.prototype.hasOwnProperty.call(args, "labels")
+}
+
+function buildLabelSelectionResponse(args = {}) {
+    const textForSuggestion = [
+        args.title,
+        args.body,
+        args.summary,
+        args.current_behavior,
+        args.actual_behavior,
+        args.expected_behavior,
+        args.error_logs,
+    ]
+        .filter((value) => typeof value === "string" && value.trim())
+        .join("\n")
+    const suggestedLabel = textForSuggestion ? inferPrimaryLabel(args.title, textForSuggestion) : null
+
+    return {
+        ok: true,
+        dry_run: true,
+        label_selection_required: true,
+        repo: `${GITHUB_OWNER}/${GITHUB_REPO}`,
+        possible_labels: DEFAULT_LABELS,
+        label_guidance: LABEL_GUIDANCE,
+        suggested_label: suggestedLabel,
+        instructions:
+            "Ask the developer to select one or more labels before gathering context, rendering the issue preview, checking duplicates, or publishing. Call create_github_issue again with labels set to the selected labels.",
+        next_action: "Wait for developer label selection, then call create_github_issue with labels.",
+    }
 }
 
 function stripExistingFooter(body) {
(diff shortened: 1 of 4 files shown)


packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+function getToken() {
+    // 1. Explicit env var
+    const envToken = process.env.GITHUB_TOKEN
+    if (envToken && envToken.trim()) {
+        return { ok: true, token: envToken.trim(), source: "env" }
+    }
+
+    // 2. GitHub CLI session — no credentials stored by Catalyst
+    try {
+        const ghToken = execSync("gh auth token", {
```
mayankmahavar1mg, line 198: not everyone has gh installed or configured

can we github api as well ?

packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+    return lines.join("\n")
+}
+
+/**
+ * Extract short search keywords from a title + body.
+ * Returns a URL-encoded query string suitable for the GitHub search API.
+ */
+function extractKeywords(title, body) {
+    // Take top words from title (most signal), pad with body words if needed
+    const stopWords = new Set([
```
mayankmahavar1mg: offload to llm 
or use standard stop keyword library

packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+        "a", "an", "the", "is", "in", "on", "at", "to", "for", "of", "and",
+        "or", "but", "not", "with", "this", "that", "it", "be", "are", "was",
+        "can", "how", "do", "does", "i", "my", "me", "we", "our",
+    ])
+    const tokenise = (str) =>
+        str
```
mayankmahavar1mg: offload this to llm tools
get desired output from llm tools

packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+}
+
+/**
+ * Search for existing issues or discussions that might be duplicates.
+ * Returns array of { number, title, url, state? } — max 5 results.
+ *
+ * Issues   → GitHub REST Search API  (GET /search/issues)
+ * Discussions → GitHub GraphQL search
+ */
+async function searchDuplicates(token, title, body, type) {
```
mayankmahavar1mg: again offload to llm and get the desired output 
let llm tell you if there is a duplicate issue or not
you drive what happens when duplicate issue is there

packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+        throw new Error(`Failed to fetch repository ID: ${err}`)
+    }
+    return res.body.data.repository.id
+}
+
+/**
+ * Fetch all available Discussion categories for the repo.
+ * Returns array of { id, name, emoji }.
+ */
+async function getDiscussionCategories(token) {
```
mayankmahavar1mg: why ?

packages/catalyst-core/mcp_v2/tools/github.js
```diff
@@ -0,0 +1,602 @@
...
+    }
+    return res.body.data.repository.discussionCategories.nodes
+}
+
+/**
+ * Pick the best matching category from available ones.
+ * Falls back to smart auto-detection based on body content.
+ */
+function resolveCategory(categories, requestedName, body) {
+    if (requestedName) {
```
mayankmahavar1mg: again offload to llm tool

mayankmahavar1mg (changes requested):
mutliple things needs to be consider by our mcp tool

1. get the error or feedback properly, get the context in good summariesed manner
2. what kind of bug/feature/request is , create label accordingly
3. template needs to be set on issue page
4. what to offload to llm vs what are we driving as mcp tool 
5. in each case needs to take dev to conclusive end

mayankmahavar1mg (commented):
## Review Summary

This PR adds GitHub issue creation to the MCP, and the **idea is exactly right** — collecting context, labeling, deduplication, templating, and guided LLM publishing is valuable. The problems are architectural: the workflow is split into too many tools and the intent classifier has a subtle ordering bug.

---

### 🔴 1. Tool count — 5 tools for 1 user action

The PR adds `start_github_issue_flow`, `gather_github_issue_context`, `preview_github_feedback`, `create_github_issue`, `generate_issue_markdown`. That is 5 MCP round-trips for "create a GitHub issue".

**Compare to how every other tool in this MCP works:**
- `debug_issue(symptom)` → reads project state, matches known errors, returns cause + fix. One call.
- `create_task_plan(goal, steps)` → stores the plan, returns it with next-step hints. One call.
- `check_config(project_path)` → reads config, validates, returns findings. One call.

Each tool does **real work** and packs guidance for the LLM into its response fields (`instructions`, `next_steps`, `next_action`). The LLM is guided via descriptions and response shapes — not via a separate tool call that just hands back a workflow spec.

**`start_github_issue_flow` does no real work.** It only returns a JSON blob of instructions. That is what the tool *description* is for. Every call to it is a wasted LLM round-trip before anything useful happens.

**Expected test case that should pass (currently won't):**
> User: "create an issue — webview is ignoring my error.html config on Android"
> → LLM should make **1 tool call** with the problem. It gets back: rendered preview, suggested label (`bug`), duplicate candidates, and a `dry_run: true` result showing the full issue body. LLM shows this to user, asks approval. User says yes → LLM calls again with `dry_run: false`. Issue published.

---

### 🔴 2. `start_github_issue_flow` — remove it, move guidance into description

This tool returns `when_to_use`, `approval_gates`, `next_steps`, `supported_labels`, `label_guidance`, `supported_templates`, `template_selection_rules`. All of this belongs in the **description of `create_github_issue`** and in the **response fields** of the actual working tools — same pattern as `debug_issue` which embeds `instructions` and `next_action` in its return value.

**Expected test case:**
> LLM receives user request "raise a feature request for dark mode support in the status bar hook"
> → LLM reads `create_github_issue` description and directly calls it with `dry_run: true`, `issue_template: "enhancement"`, `title`, `summary`. No warm-up call needed.

---

### 🔴 3. `gather_github_issue_context` — merge into `create_github_issue` internals

`debug_issue` already reads project context (package.json, config, git state) inline. `gather_github_issue_context` does the same thing, just as a separate tool. The context-gathering logic in `github.js` (`handle_gather_github_issue_context`) should be called internally by `create_github_issue`, not exposed as a public MCP tool the LLM must remember to call.

**Expected test case:**
> `create_github_issue({ title: "useVideoStream crashes on Android 13", dry_run: true, project_path: "/my/app" })`
> → Returns rendered issue body already enriched with: catalyst-core version, node version, git branch/commit, relevant file snippets. LLM did not need to call a separate context tool.

---

### 🟡 4. `preview_github_feedback` — replace with `dry_run: true` on `create_github_issue`

The preview step is a good pattern — show the developer what will be published before committing. But it does not need its own tool. A `dry_run: true` parameter on `create_github_issue` returns the exact same data (rendered body, labels, duplicate candidates) without publishing. This is the standard pattern (e.g. Terraform plan, npm publish --dry-run).

```js
// Instead of two separate tools, one tool handles both phases:
create_github_issue({ title, body, ..., dry_run: true })   // → preview + duplicates, no publish
create_github_issue({ title, body, ..., dry_run: false })  // → publishes
```

---

### 🟡 5. `generate_issue_markdown` — internal fallback only, not a public tool

`create_github_issue` already calls `handle_generate_issue_markdown` internally when auth fails (see line ~832 in github.js). The markdown fallback should stay internal — the LLM does not need to call it separately. Exposing it as a public tool bloats the tool list and gives the LLM an escape hatch to skip the real workflow.

---

### 🔴 6. Intent pattern ordering bug — `feedback` is shadowed by `debug`

**File:** `packages/catalyst-core/mcp_v2/mcp.js` line 102–106

```js
// Current order — BUG:
debug:    /error|fail|broken|not work|crash|issue|bug|why|wrong/i,  // matches "report a bug", "create an issue"
feedback: /\b(issue|bug\s+report|...)\b/i,                          // never reached for those phrases
```

`classifyIntent` returns `intents[0]` — first match wins. Because `debug` is checked before `feedback`, phrases like `"report a bug"` or `"create an issue"` match `debug` and the issue workflow is never triggered.

**Fix:** move `feedback` before `debug`.

```js
// Correct order:
feedback: /\b(issue|bug\s+report|report\s+(a\s+)?bug|open\s+(an?\s+)?issue|create\s+(an?\s+)?issue|raise\s+(an?\s+)?issue|discussion|discuss|feature\s+request|proposal|suggest)\b/i,
debug:    /error|fail|broken|not work|crash|issue|bug|why|wrong/i,
```

**Test cases that must pass after fix:**
| Query | Expected intent |
|---|---|
| `"report a bug with useVideoStream"` | `feedback` |
| `"create an issue for the error.html fallback"` | `feedback` |
| `"raise a feature request"` | `feedback` |
| `"why is my build failing"` | `debug` |
| `"error in Android WebView"` | `debug` |

---

### 🟡 7. `githubRequest` — missing HTTP timeout (from CodeRabbit scan)

**File:** `packages/catalyst-core/mcp_v2/tools/github.js` line 202

No timeout is set on the `https.request`. If GitHub is unresponsive, the MCP tool call hangs indefinitely and blocks the LLM. `runGit` in the same file already uses `timeout: 4000` as precedent.

```js
// Add to req object:
req.setTimeout(30000, () => {
    req.destroy(new Error("GitHub API request timed out after 30 seconds"))
})
```

**Test case:** GitHub API unreachable → tool returns error within 30s, LLM can fall back to markdown. Currently: hangs forever.

---

### Suggested final tool surface

| Tool | Purpose |
|---|---|
| `create_github_issue` | Does everything: gathers context, dedupes, templates, previews (`dry_run:true`), publishes (`dry_run:false`), falls back to markdown on auth failure |

That matches the single-tool pattern of every other tool in this MCP.

mayankmahavar1mg (commented):
test

mayankmahavar1mg:
## CodeRabbit CLI Review

| # | Severity | File | Location | Issue |
|---|----------|------|----------|-------|
| 1 | 🔴 Major | `packages/catalyst-core/mcp_v2/tools/github.js` | L454, L625 | Path containment check uses `absolute.startsWith(root)`, bypassable via sibling directories (e.g. `project-secrets` passes when root is `project`). Should use `path.relative(root, candidate)` and reject `..`/absolute results. Affects both `normalizeImages` and `collectRelatedFiles`. |
| 2 | 🔴 Major | `packages/catalyst-core/mcp_v2/tools/github.js` | L490-493 | `handle_create_github_issue` always includes `gatherGithubIssueContext` (git branch/commit/status, config.json, package.json, source snippets) and `enrichBody` (local filesystem project path) in the public issue body with no opt-in flag — leaks local dev environment details by default. |
| 3 | 🔴 Major | `packages/catalyst-core/mcp_v2/tools/github.js` | L763 | Empty-body validation (`!built.body`) runs *after* auto-gathered context is merged in, so a call with only `title` + `labels` can still pass validation since context filler makes the body non-empty. Should validate raw user-provided fields before merging generated context. |
| 4 | 🔴 Major | `packages/catalyst-core/mcp_v2/tools/github.js` | L138-145 | `safeReadText` reads the entire file into memory via `fs.readFileSync` before truncating to `maxBytes` — doesn't bound actual I/O/memory use. Use a bounded `fs.readSync` instead. |
| 5 | 🟡 Minor | `packages/catalyst-core/mcp_v2/tools/github.js` | L321-325 | `stripExistingFooter` regex uses the `/m` flag, so `$` can match end-of-line anywhere rather than end-of-string, risking incorrect stripping of a footer-like block from the middle of a report. |
| 6 | 🟡 Minor | `packages/catalyst-core/mcp_v2/tools/github.js` | L495-497 | Bug-template path uses `appendSection` directly for `environment`, dropping object-valued environment data instead of JSON-stringifying it like the non-bug branch does. |
| 7 | 🟡 Minor | `packages/catalyst-core/mcp_v2/mcp.js` | L459-465 | `steps_to_reproduce` schema uses `oneOf` with mixed `string`/`array` types, which can break MCP/function-calling clients that don't support union schemas. Flatten to a single type and normalize in code. |

*Generated via `coderabbit review --agent --base main`, verified against PR branch `update_issue_discussion_mcp_server`.*

PR #323, opened 2026-07-21
fix: eliminate SSR streaming race between onShellReady and onAllReady

### LINKED ISSUE #320: SSR streaming race: pipe(res) in onShellReady competes with res.write()/res.end() in onAllReady, truncating every SSR response
## Summary

The SSR renderer's use of `renderToPipeableStream` has a completion race that corrupts every streamed document. `onShellReady` pipes React's output directly into the Express response (`pipe(res)`), while `onAllReady` independently writes the deferred asset `<script>` tags and calls `res.end()`. These are two unsynchronized "finish the response" paths writing to the same stream, and one of them always loses:

- On react-dom 19.x, for typical documents, `onAllReady` wins: the handler ends `res` first, and React's own closing `</body></html>` chunk (plus any pending Suspense boundary patch-up chunk) is dropped with `ERR_STREAM_WRITE_AFTER_END`. Every page is served without `</body></html>`; pages with an async `<Suspense>` boundary are served with the fallback permanently baked into the SSR HTML.
- With other document compositions (and typically on react-dom 18.x), React wins: `pipe(res)` auto-ends the response when the stream completes, and `onAllReady`'s asset writes land on an ended response, silently dropping the deferred bundle `<script>` tags instead.

Node swallows the write-after-end when no `error` listener is attached to `res`, so nothing crashes and nothing logs. The client just receives a truncated document. This reproduces deterministically on react-dom `19.0.0` (the version catalyst-core pins) and `19.2.7` (current latest), and intermittently on `18.3.1`.

First noticed in a downstream app pinned to `catalyst-core@0.2.0-beta.2` (whose dist handler wrote `firstFoldCss`/`firstFoldJS` in `onAllReady`, same shape); confirmed still present on current `main`.

## Where the bug lives

[`packages/catalyst-core/src/server/renderer/handler.jsx#L244-L291`](https://github.com/tata1mg/catalyst-core/blob/d5721873d43d9752cb558d34e20726d0702ade35/packages/catalyst-core/src/server/renderer/handler.jsx#L244-L291) (`_renderMarkUp`):

```js
return new Promise((resolve, reject) => {
    const { pipe } = renderToPipeableStream(<CompleteDocument />, {
        onShellReady() {
            res.setHeader("content-type", "text/html")
            pipe(res)                       // (1) React now owns res and will end it
        },
        onAllReady() {
            // ...
            res.write(`<script>window.__CATALYST_IS_BOT__=...</script>`)  // (2) races (1)
            res.write(`<script>window.__SSR_RENDERED_COMPONENTS__=...`)
            // ...deferred CSS/JS writes...
            res.end()                       // (3) or (1) already ended res, or this
                                            //     truncates React's remaining output
        },
        ...
```

## Why these two callbacks compete (React's documented contract and internals)

Per the official docs ([react.dev: renderToPipeableStream](https://react.dev/reference/react-dom/server/renderToPipeableStream)), `onShellReady` and `onAllReady` are two alternative *pipe points*, not a "start" and "finish" pair:

> `onAllReady`: A callback that fires when all rendering is complete, including both the shell and all additional content. You can use this instead of `onShellReady` for crawlers and static generation. If you start streaming here, you won't get any progressive loading. The stream will contain the final HTML.

The documented crawler pattern pipes in `onShellReady` **or** in `onAllReady`, never both, and no documented pattern writes to the destination after calling `pipe()` or ends it manually. Asset injection is React's job via `bootstrapScripts` / `bootstrapScriptContent` / `bootstrapModules`:

> React will inject the doctype and your bootstrap `<script>` tags into the resulting HTML stream.

The internals make the failure on React 19 deterministic rather than timing-dependent. In `react-dom-server.node.production.js` (19.2.7):

- `completeAll()` calls `request.onAllReady()` synchronously inside the work loop the moment `allPendingTasks` reaches 0 (around L6116-6127), before the final flush of remaining data.
- The closing `</body></html>` tags are a **postamble React itself writes during the final flush**: in `flushCompletedQueues`'s `finally` block (around L7001-7015) it writes `endChunkForTag("body")` and `endChunkForTag("html")`, then calls `destination.end()`. Flushes are scheduled with `setImmediate` (`enqueueFlush`, around L7036-7047).

So on 19.x, `onAllReady` always fires before React has written the document's closing tags. A handler that calls `res.end()` inside `onAllReady` guarantees React's postamble hits a closed stream. (On 18.x the closing tags were part of the root segment rather than a postamble, which is why the race direction there flips: React's `pipe(res)` tends to end `res` first and the handler's asset writes are what get dropped. Same defect, opposite casualty.)

Note this also means the handler's current intent (asset tags before `</body>`) is not what ships even when the writes survive: they land after `</html>`.

## Reproduction 1: minimal standalone script

Mirrors the handler's exact stream wiring, nothing else. Any Node >= 18 (verified on Node 24.15.0, macOS):

```bash
mkdir repro && cd repro
npm init -y && npm install react@19.2.7 react-dom@19.2.7
node repro.mjs
```

```js
// repro.mjs - mirrors the stream wiring of
// packages/catalyst-core/src/server/renderer/handler.jsx (renderMarkUp)
import http from "node:http"
import { createElement as h, Suspense } from "react"
import { renderToPipeableStream } from "react-dom/server"

let gate
function Delayed() {
    if (!gate.done) throw gate.promise
    return h("p", null, "streamed-in content")
}

const Page = () =>
    h(
        "html",
        { lang: "en" },
        h("head", null, h("title", null, "repro")),
        h(
            "body",
            null,
            h("div", { id: "app" }, "hello"),
            h(Suspense, { fallback: h("p", null, "loading...") }, h(Delayed))
        )
    )

const server = http.createServer((req, res) => {
    gate = { done: false }
    gate.promise = new Promise((r) => setTimeout(() => ((gate.done = true), r()), 50))
    res.setHeader("content-type", "text/html; charset=utf-8")
    res.on("error", (e) => console.error("res 'error' event:", e.code))

    const { pipe } = renderToPipeableStream(h(Page), {
        onShellReady() {
            pipe(res) // handler.jsx: pipe straight into res
        },
        onAllReady() {
            // handler.jsx: append asset tags, then end the response
            res.write(`<script>window.__CATALYST_IS_BOT__=false;</script>`)
            res.write(`<script async src="/assets/app.js"></script>`)
            res.end()
        },
        onError(error) {
            console.error("onError:", String(error))
        },
    })
})

server.listen(0, () => {
    http.get({ port: server.address().port }, (r) => {
        let body = ""
        r.on("data", (c) => (body += c))
        r.on("end", () => {
            console.log("--- response body ---")
            console.log(body)
            console.log("---------------------")
            console.log("closing </html> present?  ", body.includes("</html>"))
            console.log("suspense content present? ", body.includes("streamed-in content"))
            process.exit(0)
        })
    })
})
```

Verbatim output on react-dom 19.2.7:

```
res 'error' event: ERR_STREAM_WRITE_AFTER_END
res 'error' event: ERR_STREAM_WRITE_AFTER_END
--- response body ---
<!DOCTYPE html><html lang="en"><head><title>repro</title></head><body><div id="app">hello</div><!--$?--><template id="B:0"></template><p>loading...</p><!--/$--><script id="_R_">requestAnimationFrame(function(){$RT=performance.now()});</script><script>window.__CATALYST_IS_BOT__=false;</script><script async src="/assets/app.js"></script>
---------------------
closing </html> present?   false
suspense content present?  false
```

The document is missing `</body></html>`, and the Suspense boundary's resolved content never reaches the client: the SSR HTML permanently shows `loading...` (the `$RC` patch-up chunk React streams to replace the fallback was written after the handler's `res.end()`).

To see the opposite race direction on the *same React version*, delete the `<head>` and the Suspense child, leaving a bare `html > body > div` page:

```
res error: ERR_STREAM_WRITE_AFTER_END
" html><html lang=\"en\"><head></head><body><div id=\"app\">hello</div></body></html>"
has </html>? true
```

Here React's flush completed and auto-ended `res` inside `pipe()` before `onAllReady` ran, so the closing tags survive but the asset `<script>` tag is silently dropped instead. Which side loses depends on document composition and version; something is always lost.

Remove the `res.on("error")` listener and both variants produce no output at all while still truncating the response: this failure is silent in production.

## Reproduction 2: this repo's own fixture app, end to end

Using the repo's normal fixture flow (same steps as `scripts/test-catalyst-core.sh`) at current main (`d572187`):

```bash
npm ci
npm run core:build
PACK_DIR=$(mktemp -d)
npm pack --workspace packages/catalyst-core --pack-destination "$PACK_DIR" --ignore-scripts
cd apps/catalyst-core-test
npm install --no-save --package-lock=false "$PACK_DIR"/catalyst-core-*.tgz
npm run build
NODE_ENV=production npm run serve
```

Then (note: the fixture server binds IPv6 localhost only on macOS):

```bash
curl -s "http://[::1]:3005/" -A "Mozilla/5.0 (Macintosh) Chrome/126" | grep -c '</html>'
```

Observed result: `0`. The 4086-byte response contains the doctype, the full app markup, the head `<script src>` tags, and the `__CATALYST_IS_BOT__` / `__SSR_RENDERED_COMPONENTS__` inline scripts, then stops. React's closing `</body></html>` chunk was written after the handler's `res.end()` and dropped. Every request does this.

## Everything else tried (for vetting)

To rule out coincidence, I ran a 27-case-per-version scenario matrix: 3 stream wirings x 9 scenarios x react-dom {18.3.1, 19.0.0, 19.2.7}. The three wirings:

- `current`: the exact shape of `handler.jsx` on main (pipe in `onShellReady`, write tail + end in `onAllReady`).
- `patch`: a control implementation that routes React through a `PassThrough` piped with `{ end: false }` and appends the tail on the PassThrough's `end` event (single completion signal, manual `res.end()`).
- `transform`: a second control, a `Transform` whose `flush()` appends the tail inside the pipe chain with auto-end preserved.

Scenarios: sync page (no Suspense), fast boundary (~10ms), slow boundary (~300ms), multiple sibling boundaries resolving out of order, nested boundaries, boundary that throws after resolving, shell render error, client disconnect mid-stream, slow client (reader paused 250ms, backpressure).

Results, identical across 19.0.0 and 19.2.7:

| wiring | write-after-end errors | missing `</html>` | Suspense content dropped |
|---|---|---|---|
| `current` (main) | every successful-render scenario (1-2 per request) | every scenario | single-boundary pages always; multi-boundary pages lose the last boundary's patch chunk |
| `patch` (control) | 0 | none | none |
| `transform` (control) | 0 | none | none |

On 18.3.1 the `current` wiring fails the same checks in the matrix, with the race direction dependent on timing (single runs occasionally produce an intact document; the downstream 0.2.0-beta.2 sighting was the tail-dropped direction).

Both control wirings passing every scenario on every version, while only the two-completion-paths wiring fails, isolates the defect to the callback usage rather than to Express, page content, or React version quirks.

Additional observations from the matrix runs:

- Client disconnect mid-stream: React fires `onError("The destination stream closed early.")` and recovers; no crash. But the handler never calls the `abort()` returned by `renderToPipeableStream`, so render work for still-pending boundaries continues after disconnect.
- The handler registers no `onShellError`; a root render error currently rejects the promise and relies on the outer catch, worth checking separately.
- When the tail writes do survive (React-wins direction on 18.x), they land after `</html>`, not before `</body>` as the code reads.

<details>
<summary>Scenario matrix runner (self-contained, for vetting)</summary>

Create `v18/`, `v19/`, `v19_2/` dirs next to the script, each with a `node_modules` containing the matching react + react-dom (e.g. `cd v19_2 && npm init -y && npm i react@19.2.7 react-dom@19.2.7`), then `node scenario-matrix.mjs`.

```js
// Scenario matrix for catalyst-core's renderToPipeableStream usage.
//
// Exercises the exact streaming shapes from
// packages/catalyst-core/src/server/renderer/handler.jsx across:
//   MODES      x  SCENARIOS  x  react-dom versions (via cwd node_modules)
//
// Driver mode (no env): spawns itself for each version dir x mode x scenario.
// Worker mode (SCENARIO/MODE env set): runs one case, prints JSON result.

import http from "node:http"
import { spawnSync } from "node:child_process"
import { fileURLToPath } from "node:url"
import path from "node:path"

const SELF = fileURLToPath(import.meta.url)

// ---------------------------------------------------------------- worker ---
async function runWorker() {
    const { createElement: h, Suspense } = await import("react")
    const { renderToPipeableStream } = await import("react-dom/server")
    const { PassThrough, Transform } = await import("node:stream")
    const reactDomVersion = (await import("react-dom/package.json", { with: { type: "json" } })).default.version

    const scenario = process.env.SCENARIO
    const mode = process.env.MODE

    const TAIL =
        `<script>window.__CATALYST_IS_BOT__=false;</script>` +
        `<script>window.__SSR_RENDERED_COMPONENTS__=new Set([])</script>` +
        `<script async src="/assets/app.deadbeef.js"></script>`

    // -- suspense helper: throws a promise until it resolves after `ms`
    const makeDelayed = (ms, { fail = false, label = "delayed" } = {}) => {
        const cell = { done: false, err: null }
        cell.promise = new Promise((resolve) =>
            setTimeout(() => {
                cell.done = true
                if (fail) cell.err = new Error(`boom in ${label}`)
                resolve()
            }, ms)
        )
        return function Delayed() {
            if (!cell.done) throw cell.promise
            if (cell.err) throw cell.err
            return h("p", { className: "resolved" }, `${label} resolved`)
        }
    }

    const Thrower = () => {
        throw new Error("shell render crash")
    }

    const buildPage = () => {
        const kids = [h("div", { id: "app" }, h("h1", null, "hello"))]
        const S = (el, key) => h(Suspense, { fallback: h("p", null, "loading..."), key }, el)
        switch (scenario) {
            case "sync":
                break
            case "suspense-fast":
                kids.push(S(h(makeDelayed(10, { label: "fast" })), "f"))
                break
            case "suspense-slow":
                kids.push(S(h(makeDelayed(300, { label: "slow" })), "s"))
                break
            case "suspense-multi":
                kids.push(S(h(makeDelayed(40, { label: "a" })), "a"))
                kids.push(S(h(makeDelayed(140, { label: "b" })), "b"))
                break
            case "suspense-nested":
                kids.push(
                    S(
                        h("div", null, h(makeDelayed(40, { label: "outer" })), S(h(makeDelayed(120, { label: "inner" })), "i")),
                        "o"
                    )
                )
                break
            case "boundary-error":
                kids.push(S(h(makeDelayed(50, { fail: true, label: "failing" })), "e"))
                break
            case "shell-error":
                kids.push(h(Thrower))
                break
            case "client-abort":
            case "slow-client":
                kids.push(S(h(makeDelayed(150, { label: "late" })), "l"))
                break
            default:
                throw new Error(`unknown scenario ${scenario}`)
        }
        return h("html", { lang: "en" }, h("head", null, h("title", null, "m")), h("body", null, ...kids))
    }

    const result = {
        scenario,
        mode,
        reactDom: reactDomVersion,
        events: [],
        writeAfterEnd: 0,
        otherResErrors: [],
        onErrorCalls: [],
        crashed: null,
        clientStatus: null,
        body: null,
        timedOut: false,
    }

    process.on("uncaughtException", (err) => {
        result.crashed = String(err && (err.code || err.message))
        conclude()
    })

    let concluded = false
    const conclude = () => {
        if (concluded) return
        concluded = true
        const b = result.body || ""
        result.checks = {
            doctype: b.startsWith("<!DOCTYPE html>"),
            closingHtml: b.includes("</html>"),
            tailScripts: b.includes("app.deadbeef.js"),
            tailAfterClosingHtml: b.includes("</html>") && b.indexOf("app.deadbeef.js") > b.indexOf("</html>"),
            suspenseResolvedContent: b.includes("resolved"),
            suspenseClientPatchScript: b.includes("$RC") || b.includes("<div hidden"),
        }
        delete result.body
        console.log(JSON.stringify(result))
        server.close()
        process.exit(0)
    }

    const server = http.createServer((req, res) => {
        res.setHeader("content-type", "text/html; charset=utf-8")
        res.on("error", (err) => {
            if (err.code === "ERR_STREAM_WRITE_AFTER_END") result.writeAfterEnd++
            else result.otherResErrors.push(String(err.code || err.message))
        })

        const writeTail = (w) => {
            w.write(TAIL)
        }

        const { pipe } = renderToPipeableStream(buildPage(), {
            onShellReady() {
                result.events.push("onShellReady")
                if (mode === "current") {
                    pipe(res)
                } else if (mode === "patch") {
                    const bundleStream = new PassThrough()
                    let flushed = false
                    const flushAssets = () => {
                        if (flushed) return
                        flushed = true
                        writeTail(res)
                        res.end()
                    }
                    bundleStream.on("error", (err) => {
                        flushed = true
                        if (!res.destroyed) res.destroy(err)
                    })
                    res.on("close", () => {
                        if (!bundleStream.destroyed) bundleStream.destroy()
                    })
                    bundleStream.on("end", flushAssets)
                    bundleStream.pipe(res, { end: false })
                    pipe(bundleStream)
                } else if (mode === "transform") {
                    const tail = new Transform({
                        transform(chunk, _enc, cb) {
                            cb(null, chunk)
                        },
                        flush(cb) {
                            this.push(TAIL)
                            cb()
                        },
                    })
                    tail.on("error", (err) => {
                        if (!res.destroyed) res.destroy(err)
                    })
                    res.on("close", () => {
                        if (!res.writableEnded && !tail.destroyed) tail.destroy()
                    })
                    tail.pipe(res) // auto-end preserved: res.end() fires when tail ends
                    pipe(tail)
                }
            },
            onAllReady() {
                result.events.push("onAllReady")
                if (mode === "current") {
                    writeTail(res)
                    res.end()
                }
            },
            onShellError(error) {
                result.events.push("onShellError")
                result.onErrorCalls.push(`shell:${error.message}`)
                if (!res.headersSent) res.writeHead(500)
                res.end("<!-- shell error -->")
            },
            onError(error) {
                result.onErrorCalls.push(String(error.message || error))
            },
        })
    })

    server.listen(0, () => {
        const { port } = server.address()
        const reqOpts = { port, host: "127.0.0.1", path: "/" }
        const creq = http.get(reqOpts, (cres) => {
            result.clientStatus = cres.statusCode
            let body = ""
            if (scenario === "slow-client") {
                cres.pause()
                setTimeout(() => cres.resume(), 250)
            }
            cres.on("data", (c) => {
                body += c
                if (scenario === "client-abort" && body.length > 0) creq.destroy()
            })
            cres.on("end", () => {
                result.body = body
                conclude()
            })
            cres.on("close", () => {
                result.body = body
                setTimeout(conclude, 300) // let post-abort server-side errors surface
            })
        })
        creq.on("error", () => setTimeout(conclude, 300))
        setTimeout(() => {
            result.timedOut = true
            conclude()
        }, 3000)
    })
}

// ---------------------------------------------------------------- driver ---
function runDriver() {
    const scratchDir = path.dirname(SELF)
    const versions = [
        { name: "react-dom 18.3.1", cwd: path.join(scratchDir, "v18") },
        { name: "react-dom 19.0.0", cwd: path.join(scratchDir, "v19") },
        { name: "react-dom 19.2.7", cwd: path.join(scratchDir, "v19_2") },
    ]
    const modes = ["current", "patch", "transform"]
    const scenarios = [
        "sync",
        "suspense-fast",
        "suspense-slow",
        "suspense-multi",
        "suspense-nested",
        "boundary-error",
        "shell-error",
        "client-abort",
        "slow-client",
    ]

    const rows = []
    for (const v of versions) {
        for (const mode of modes) {
            for (const scenario of scenarios) {
                const r = spawnSync(process.execPath, [SELF], {
                    cwd: v.cwd,
                    env: { ...process.env, SCENARIO: scenario, MODE: mode },
                    encoding: "utf-8",
                    timeout: 10000,
                })
                let parsed
                try {
                    parsed = JSON.parse(r.stdout.trim().split("\n").pop())
                } catch {
                    parsed = { scenario, mode, crashed: `no-json (stderr: ${String(r.stderr).slice(0, 120)})` }
                }
                parsed.version = v.name
                rows.push(parsed)
            }
        }
    }

    // ---- render summary table
    const flag = (ok) => (ok ? " ok " : "FAIL")
    console.log(
        "version           | mode      | scenario        | wAfterEnd | crash | doctype | </html> | tail | suspense | onError"
    )
    console.log("-".repeat(120))
    for (const r of rows) {
        const c = r.checks || {}
        const suspenseRelevant = !["sync", "shell-error"].includes(r.scenario)
        console.log(
            [
                r.version.padEnd(17),
                r.mode.padEnd(9),
                r.scenario.padEnd(15),
                String(r.writeAfterEnd ?? "?").padStart(9),
                (r.crashed ? "CRASH" : "  -  ").padEnd(5),
                (r.scenario === "shell-error" ? "  n/a " : flag(c.doctype)).padEnd(7),
                (r.scenario === "shell-error" ? " n/a " : flag(c.closingHtml)).padEnd(7),
                (["shell-error", "client-abort"].includes(r.scenario) ? " n/a" : flag(c.tailScripts)).padEnd(4),
                (suspenseRelevant && !["client-abort"].includes(r.scenario)
                    ? flag(c.suspenseResolvedContent)
                    : " n/a ").padEnd(8),
                (r.onErrorCalls || []).join(";").slice(0, 40),
            ].join(" | ")
        )
    }
    const fs = requireFs()
    fs.writeFileSync(path.join(scratchDir, "matrix-results.json"), JSON.stringify(rows, null, 2))
    console.log("\nfull results: matrix-results.json")
}

function requireFs() {
    return process.getBuiltinModule ? process.getBuiltinModule("fs") : require("fs")
}

if (process.env.SCENARIO) runWorker()
else runDriver()
```

</details>

## Impact

- Every SSR response from the current handler is malformed HTML (missing `</body></html>` on react-dom 19, missing deferred bundle scripts in the other race direction).
- Pages with an async Suspense boundary serve the fallback in the SSR HTML permanently; the streamed patch-up never arrives. Bots and no-JS clients never see the real content; hydrating clients recover only by client-rendering it, defeating streaming SSR.
- In the tail-dropped direction, the client bundle scripts never reach the browser at all, so nothing hydrates.
- The failure emits no log by default (Node write-after-end without an `error` listener is silent), so it ships unnoticed.

## Ruled out

- Express/middleware interference: reproduces on a bare `node:http` server.
- Page content/app specifics: reproduces with a three-element document.
- React version regression: reproduces on 18.3.1, 19.0.0, and 19.2.7; only the failure direction varies.
- The two control wirings (single completion signal) pass all 27 cases on every version, so the defect is specifically the `pipe(res)`-in-`onShellReady` + write/end-in-`onAllReady` combination.

## Environment

- catalyst-core `main` @ `d572187` (0.3.0-beta.1) and dist `0.2.0-beta.2`
- react / react-dom 18.3.1, 19.0.0, 19.2.7
- Node v24.15.0, macOS (Darwin 25.4.0); fixture app also verified via `catalyst serve` as described above

Uttkarsh-Srivastava:
## Summary
- `renderToPipeableStream`'s `pipe(res)` in `onShellReady` and the handler's own `res.write(firstFoldCss/firstFoldJS)`/`res.end()` in `onAllReady` (`packages/catalyst-core/src/server/renderer/handler.js`) are two independent completion paths racing to close the same response. Whichever loses is silently dropped (no `error` listener on `res`), truncating every SSR response and permanently sticking async `<Suspense>` boundaries on their fallback in some documents.
- Fixes this by routing React's output through an intermediate `Transform` ("tail"). React auto-ends `tail` exactly like it auto-ended `res` before, but `tail`'s `flush()` hook is guaranteed to run exactly once — appending `firstFoldCss`/`firstFoldJS` — before a plain `tail.pipe(res)` is allowed to end `res`. Exactly one caller of `res.end()` now.
- Progressive streaming is unchanged: every chunk still flows to `res` immediately as React writes it; only the trailing asset tags were ever deferred to the end, same as before.

Fixes #320

Visual walkthrough of the race and the fix: https://claude.ai/code/artifact/79a44698-079d-47f0-b72e-911a352f05e4

## Test plan
- [ ] Build a fixture app against this branch and confirm `</html>` is present and Suspense content resolves
- [ ] Verify `firstFoldCss`/`firstFoldJS` still injected correctly for both bot and non-bot requests

### DIFF
--- a/packages/catalyst-core/src/server/renderer/handler.jsx
+++ b/packages/catalyst-core/src/server/renderer/handler.jsx
@@ -23,6 +23,7 @@ import {
     generateModulePreloadLinkElements,
 } from "./extract.js"
 import path from "path"
+import { Transform } from "node:stream"
 
 import CustomDocument from "@catalyst/template/server/document"
 
@@ -242,13 +243,17 @@ const _renderMarkUp = async (
         res.status(status)
 
         return new Promise((resolve, reject) => {
-            const { pipe } = renderToPipeableStream(<CompleteDocument />, {
-                onShellReady() {
-                    res.setHeader("content-type", "text/html")
-                    pipe(res)
+            // Single completion path: React's pipe() auto-ends `tail`, and
+            // `flush()` appends the deferred asset tags before that end
+            // signal propagates to `res` via the plain pipe below. This
+            // avoids the onShellReady/onAllReady race where React's own
+            // stream-end and a manual res.end() compete to close `res`
+            // (see issue #320).
+            const tail = new Transform({
+                transform(chunk, _enc, cb) {
+                    cb(null, chunk)
                 },
-
-                onAllReady() {
+                flush(cb) {
                     // Deferred assets — injected after body (non-blocking)
                     const deferredAssets = chunkExtractor
                         ? chunkExtractor.getDeferredAssets()
@@ -256,10 +261,10 @@ const _renderMarkUp = async (
 
                     // Tell client which components were SSR'd so split() can
                     // eagerly import them (prevents Suspense fallback flash)
-                    res.write(`<script>window.__CATALYST_IS_BOT__=${isBot ? "true" : "false"};</script>`)
+                    this.push(`<script>window.__CATALYST_IS_BOT__=${isBot ? "true" : "false"};</script>`)
                     if (chunkExtractor) {
                         const renderedKeys = chunkExtractor.getRenderedComponentKeys()
-                        res.write(
+                        this.push(
                             `<script>window.__SSR_RENDERED_COMPONENTS__=new Set(${JSON.stringify(renderedKeys)})</script>`
                         )
                     }
@@ -270,13 +275,24 @@ const _renderMarkUp = async (
                         isBot
                     )
                     if (newCssPaths.length) {
-                        res.write(`<style>${readCssFromDisk(newCssPaths, buildDir)}</style>`)
+                        this.push(`<style>${readCssFromDisk(newCssPaths, buildDir)}</style>`)
                     }
                     if (!isBot) {
-                        res.write(generateScriptStrings(deferredAssets.js))
+                        this.push(generateScriptStrings(deferredAssets.js))
                     }
 
-                    res.end()
+                    cb()
+                },
+            })
+            tail.pipe(res)
+
+            const { pipe } = renderToPipeableStream(<CompleteDocument />, {
+                onShellReady() {
+                    res.setHeader("content-type", "text/html")
+                    pipe(tail)
+                },
+
+                onAllReady() {
                     cleanupSafeArea()
                     resolve()
                 },
@@ -285,6 +301,7 @@ const _renderMarkUp = async (
                     console.error("Error in renderToPipeableStream:", error)
                     safeCall(onRenderError, { req, res, store, error })
                     cleanupSafeArea()
+                    tail.destroy(error)
                     reject(error)
                 },
             })


mayankmahavar1mg:
## Code Review — CodeRabbit + Manual Analysis

**CodeRabbit CLI**: 0 automated findings on the changed file (`handler.jsx`). Additional manual review of the diff and surrounding error-handling paths below.

| # | Severity | Location | Finding | Status |
|---|----------|----------|---------|--------|
| 1 | Info | `handler.jsx:284` | `onShellError` still isn't implemented — shell-level render errors fall through to the generic `onError`. This is a **pre-existing gap**, not introduced by this PR (confirmed against `main`). This PR actually improves that path: `tail.destroy(error)` now properly tears down `res` instead of leaving the connection hanging (previously `onError` never touched `res` at all). Non-blocking; worth a follow-up. | Out of scope |

### What this PR does well
- Correctly identifies and fixes a real race: the old code had two independent completion signals (`pipe(res)` implicitly ending `res`, and a manual `res.end()` inside `onAllReady`) that could compete to close `res`.
- The `Transform`-based `tail` buffer is a clean single-completion-path fix — `flush()` injects the deferred asset tags before Node's `.pipe()` auto-ends `res`, guaranteeing exactly one completion signal.
- Bonus fix: `onError` now calls `tail.destroy(error)`, correctly propagating and closing `res` on error — previously `res` was never terminated on error, which could hang the connection.
- Clear, accurate inline comment explaining the fix and referencing issue #320.

### Rating: ⭐⭐⭐⭐⭐⭐⭐⭐⭐☆ (9/10)

Solid, well-scoped bug fix with a correct mechanism and a helpful explanatory comment. Docked one star only because `onShellError` remains unhandled (pre-existing, arguably out of scope here, but a natural companion fix given the theme).

---
*Reviewed with [CodeRabbit CLI](https://www.coderabbit.ai/cli) + manual analysis.*

PR #284, opened 2026-06-24
Vite Release

mayansharma:
**PR Summary**

Migrates Catalyst Core’s web runtime from the old webpack/@tata1mg/router stack to a Vite-first SSR pipeline, with updated routing, split rendering, asset extraction, template support, and test coverage.

**Core Runtime**
- Replaces webpack SSR/client build flow with Vite server/client builds.
- Adds Vite configs for shared, client, and server builds.
- Adds a custom Node ESM loader for aliases, extension resolution, JSX, CJS interop, and legacy `require()` compatibility.
- Moves Catalyst entrypoint to `index.jsx` and re-exports React Router plus Catalyst web-router helpers.
- Removes old webpack Babel configs, remove-client/remove-ssr plugins, and legacy webpack server startup paths.
- Adds `preServerInit` to preload app config/hooks before Express starts.

**SSR, Routing, And Data**
- Adds Catalyst-owned web-router layer: `RouterDataProvider`, `MetaTag`, route data fetching, metadata merging, router context, and typed declarations.
- Updates `ClientRouter` and `ServerRouter` to use the new prepared route structure.
- Fixes production route matching to use `req.originalUrl`.
- Adds optional store fallback so apps without Redux store files still run.
- Moves renderer files to JSX and splits document rendering into `Head`, `Body`, `render`, and `handler`.

**Vite Split And Asset Pipeline**
- Adds `split()` and `hydrationReady()` as the Vite replacement for `@loadable/component`.
- Adds `SplitInview` for visibility-gated loading and per-instance fallback support.
- Adds cache-key injection for `split()` calls.
- Adds categorized asset manifest generation for `ssrTrue`, `ssrFalse`, `essential`, and orphan chunks.
- Adds chunk extraction for critical/deferred JS and CSS.
- Preloads critical assets, inlines critical CSS, and scopes deferred asset caching by route.
- Fixes split statics propagation so `serverFetcher`, `clientFetcher`, and `setMetaData` remain available after lazy loading.
- Fixes production static serving for Vite assets under `PUBLIC_STATIC_ASSET_PATH/client/assets`.

**Server Runtime**
- Reworks Express server startup for Vite SSR.
- Preloads production render module once at startup.
- Adds build manifest cache so manifests are read once instead of per request.
- Adds CJS bridge helper for ESM server files.
- Supports custom Vite config loading.
- Adds `serve:inspect`.
- Improves env loading and client env define handling.

**Templates And Fixture App**
[...]

PR #326, opened 2026-07-29
fix: patch @tailwindcss/postcss base so Vite scans the whole app

### LINKED ISSUE #325: Vite migration: Tailwind v4 content scanner doesn't detect classes when CSS entry is outside component tree
## Problem

- After the Vite migration (#284), Tailwind v4 stops picking up utility classes used in app components.
- The generated CSS only has Tailwind's default boilerplate (`.flex`, `.hidden`, `.container`, `.ring`, etc.) — classes actually used in JSX (`bg-[var(--bg)]`, `gap-3`, `rounded-xl`, `px-3`, ...) never make it into the output.
- Page renders and hydrates fine — this is a styling-only regression, not a crash.
- `useai` was built and ran correctly on the old webpack build (created 2026-06-30, with Tailwind from day one). The Vite migration (#284, merged 2026-07-15) landed afterward and broke it — this is a genuine webpack → Vite regression, not a pre-existing gap.

## Why it worked under webpack but breaks under Vite

Background: Tailwind v4 has no config file. Instead of a `content` glob, `@tailwindcss/postcss` auto-scans for class names starting from a **scan root**, computed as `dirname(from)` where `from` is the path of the file currently being compiled — unless the plugin is given an explicit `base` option to override this. catalyst-core's convention puts the app's global stylesheet at `src/static/css/base/` (via the `@css` alias) while components live under `src/js/**` — two directory trees with no containment relationship, so where the scan root lands matters a lot here.

**Before (webpack, working):**
- Same `postcss.config.js`, same `{ "@tailwindcss/postcss": {} }`, present from this app's first commit — no `base` was ever configured, then or now.
- Webpack's `postcss-loader` invokes PostCSS in a way that does not populate a per-file `from` this scanner keys off the same way Vite does — in practice, the plugin's scan root was effectively the app root (`process.cwd()`), so it walked the entire app, including `src/js/**`.
- Result: real component classes (`gap-3`, `rounded-xl`, `bg-[var(--bg)]`, etc.) were found and compiled in. Styling worked.

**After (Vite, broken):**
- Vite's dev/build pipeline calls the PostCSS transform per-file and passes that file's real, resolved path as `from`.
- For `src/static/css/base/index.scss`, `dirname(from)` resolves to `src/static/css/base/` — a directory with no JS/JSX files in or under it.
- `packages/catalyst-core/src/vite/vite.config.js`'s `css.preprocessorOptions.scss` block configures SCSS variable injection (`getScssResourcesImport`, `font_url`, `url_for`) but never passes a `base` to the Tailwind/PostCSS plugin to compensate.
- Result: the scanner never reaches `src/js/**`, so only Tailwind's always-emitted default utilities appear in the compiled CSS — everything the app actually uses is silently dropped.

Nothing in `postcss.config.js` or the app's Tailwind setup changed between the two states — the only change was the build tool (webpack → Vite) and the resulting `from` path passed into the same PostCSS plugin.

## Repro

1. In a catalyst-core Vite app, add Tailwind v4 with no `tailwind.config.js` (v4 zero-config):
   ```js
   // postcss.config.js
   module.exports = { plugins: { "@tailwindcss/postcss": {} } }
   ```
2. Global stylesheet at the conventional path, `src/static/css/base/index.scss`:
   ```scss
   @import "styles";
   @import "tailwindcss";
   ```
3. Imported once via `client/styles.js` → `client/index.js`.
4. Use Tailwind classes in a component under `src/js/**`, e.g.:
   ```jsx
   <div className="flex gap-3 rounded-xl px-3 py-1.5 bg-[var(--bg)]">...</div>
   ```
5. `npm start`, then check the compiled CSS:
   ```
   curl http://<host>:<port>/src/static/css/base/index.scss
   ```

**Expected:** compiled CSS includes `.flex`, `.gap-3`, `.rounded-xl`, `.px-3`, `.py-1.5`, and the arbitrary-value class for `bg-[var(--bg)]`.

**Actual:** only Tailwind's default boilerplate utilities are present — none of the app's actual classes.

## Environment

- Regressed by `d572187` (Vite Release #284, merged 2026-07-15)
- Previously working on catalyst-core's webpack build (`useai` created 2026-06-30 with Tailwind from the start)
- `tailwindcss` / `@tailwindcss/postcss` v4.3.3

## Suggested fix

- Pass an explicit `base` (app root, e.g. `process.env.src_path`) to `@tailwindcss/postcss` in `packages/catalyst-core/src/vite/vite.config.js`, so the scanner isn't anchored to the CSS file's own directory.
- Framework-level fix — any app using the conventional `@css` → `src/static/css` layout with components under `src/js/**` hits this.

## Related upstream reports

This is a known class of issue in Tailwind v4's automatic source detection, not unique to catalyst-core:

- [tailwindlabs/tailwindcss#13136](https://github.com/tailwindlabs/tailwindcss/issues/13136) — monorepo case where auto-detection only scans the "direct package," missing sibling/parent directories. Workaround was the `base` option, but it only accepts absolute paths (not `../..`-style relative paths), which limits portability.
- [tailwindlabs/tailwindcss#16480](https://github.com/tailwindlabs/tailwindcss/issues/16480) — Vite + Svelte repro of the mirror problem (wrong scope, not missing scope). Maintainer-suggested fix was an explicit `@source` directive (optionally paired with `source(none)` to disable auto-detection first) — the same mechanism used as the app-level workaround here.
- [Tailwind docs: Detecting classes in source files](https://tailwindcss.com/docs/detecting-classes-in-source-files) — documents the heuristic-based auto-detection behavior and the `@source`/`base` override mechanisms.

Uttkarsh-Srivastava:
## Summary
- Tailwind v4's zero-config scanner (`@tailwindcss/postcss`) computes its scan root as `dirname(from)`. Vite passes each CSS file's own resolved path as `from`, so for catalyst-core's conventional layout (global stylesheet under `src/static/css`, components under `src/js`) the scanner never reaches component files and silently drops every class actually used in JSX. Webpack's `postcss-loader` didn't populate `from` the same way, so this only regressed after the Vite migration (#284).
- `packages/catalyst-core/src/vite/vite.config.js` now resolves the app's own `postcss.config.js` and injects `base: process.env.src_path` into `@tailwindcss/postcss` when the app hasn't set one, leaving every other plugin's options untouched. If the app's postcss config can't be patched this way (no file, or array-form plugins), `css.postcss` is left unset so Vite falls back to its normal auto-loaded resolution — no behavior change for apps without Tailwind.

Fixes #325

## Test plan
- [x] `node --check` on the modified `vite.config.js`
- [x] Standalone Node simulation with fake `@tailwindcss/postcss` + `autoprefixer` packages confirming only the Tailwind plugin gets `base` injected and other plugins pass through untouched
- [ ] Manual repro from the issue against a real Vite app (add Tailwind v4 zero-config, confirm compiled CSS includes classes used in `src/js/**`)

### DIFF
--- a/packages/catalyst-core/src/vite/vite.config.js
+++ b/packages/catalyst-core/src/vite/vite.config.js
@@ -130,6 +130,51 @@ try {
 const allAliases = { ..._moduleAliases, ...catalyst_moduleAliases }
 
 import { imageUrl, fontUrl } from "./scssParams.js"
+import { createRequire } from "module"
+
+// Tailwind v4's zero-config scanner (`@tailwindcss/postcss`) walks up from
+// `dirname(from)` — the directory of the file PostCSS is currently compiling —
+// unless given an explicit `base`. Vite passes each CSS file's own resolved
+// path as `from`, so for catalyst-core's conventional layout (global
+// stylesheet under `src/static/css`, components under `src/js`, no shared
+// parent short of the app root) the scanner never reaches component files and
+// silently drops every class actually used in JSX. Webpack's postcss-loader
+// doesn't populate `from` the same way, which is why this only regresses
+// under Vite. We resolve the app's own postcss.config.js ourselves and patch
+// in `base: process.env.src_path` so the scan root covers the whole app.
+// Only the `{ pluginName: options }` shorthand can be patched this way —
+// array-form configs (`[require(...)()]`) have already closed over their
+// options — so anything else is left alone and Vite falls back to its normal
+// auto-loaded postcss.config.js resolution.
+const resolvePostcssPlugins = () => {
+    const configPath = path.join(process.env.src_path, "postcss.config.js")
+    if (!fs.existsSync(configPath)) return null
+
+    const appRequire = createRequire(configPath)
+
+    try {
+        const userConfig = appRequire(configPath)
+        const pluginsConfig = userConfig?.plugins
+
+        if (!pluginsConfig || Array.isArray(pluginsConfig) || typeof pluginsConfig !== "object") {
+            return null
+        }
+
+        return Object.entries(pluginsConfig).map(([name, options]) => {
+            const plugin = appRequire(name)
+            const finalOptions =
+                name === "@tailwindcss/postcss" && (!options || options.base === undefined)
+                    ? { ...options, base: process.env.src_path }
+                    : options
+            return plugin(finalOptions)
+        })
+    } catch (error) {
+        console.warn(`Failed to patch postcss config at ${configPath}:`, error.message)
+        return null
+    }
+}
+
+const postcssPlugins = resolvePostcssPlugins()
 
 const alias = () => {
     if (!allAliases || typeof allAliases !== "object") {
@@ -377,6 +422,7 @@ export default defineConfig({
             localsConvention: "camelCase",
             generateScopedName: "[name]__[local]___[hash:base64:5]",
         },
+        ...(postcssPlugins ? { postcss: { plugins: postcssPlugins } } : {}),
         preprocessorOptions: {
             scss: {
                 additionalData: (content) =>


mayankmahavar1mg:
**Released in `catalyst-core@0.3.0-beta.2`**

- Fixed Tailwind v4's zero-config scanner (`@tailwindcss/postcss`) not picking up classes used in component files after the Vite migration
- `vite.config.js` now injects `base: process.env.src_path` into the Tailwind PostCSS plugin only, leaving other plugins/config untouched
- No behavior change for apps without Tailwind

Issue #359, opened 2026-08-03
Implement unified error codes (RFC #155)


mayankmahavar1mg:
## Description
Implement unified error codes per RFC #155: PREFLIGHT-xxx, CONFIG-xxx, IOS-xxx, ANDROID-xxx, BUNDLE-xxx, RUNTIME-WEB-xxx, RUNTIME-NATIVE-xxx, PROCESS-xxx. Reconcile with existing errors.js and validator.js implementations.

## Acceptance criteria
- [ ] All error categories implemented with unique codes
- [ ] Existing errors.js/validator.js usages migrated to the new code system

## Parent
Story: #343

mayankmahavar1mg:
## Proposal

Current state: two disconnected, incomplete error systems exist. `src/native/bridge/errors.js` has a real taxonomy (`ERROR_CATEGORIES`, `ERROR_CODES`, `ERROR_DEFINITIONS`, `createStandardError`, `translateError`) but only for native-bridge/device errors. Everywhere else — both `validator.js` files, `scripts/build.js`, native build scripts, `server/expressServer.js`, `server/renderer/handler.jsx` — errors are raw `new Error("string")` or ad hoc `console.error`, with no code, no category. `logger.js` (a built Winston logger) has zero callers anywhere in `src` today.

### 1. Runtime registry

`src/errors/registry.js` — one entry per catalyst-owned code:

```js
"PREFLIGHT-001": {
  category: "PREFLIGHT",
  defaultMessage: "config.json is missing required keys",
  defaultDetails: "...",
  recoverable: true,
  suggestedAction: "...",
  docUrl: "https://github.com/tata1mg/catalyst-core/blob/main/errors/PREFLIGHT/PREFLIGHT-001.md",
}
```

`docUrl` is a real GitHub blob URL so a terminal error can link straight to the explanation.

`src/errors/index.js` exports `ERROR_CODES`, `CatalystError` (extends `Error`; `{ code, category, message, details, recoverable, suggestedAction, docUrl, cause }`), `createError(code, overrides)`, and `wrapForeignError(stage, err)`.

### 2. Foreign errors — wrap, don't replace

If a caught error already has its own code (e.g. Rollup's `PLUGIN_LOAD_ERROR`), we don't invent a competing catalyst code for it. `wrapForeignError(stage, err)` assigns one generic per-stage wrapper code (e.g. `BUNDLE-000`), sets `cause: err`, and always prints the original foreign code/message verbatim:

```
[...]

PR #373, opened 2026-08-03
feat(bridge): wire WebBridge.js dispatch errors into error registry

### LINKED ISSUE #370: Wire WebBridge.js dispatch errors into the error registry
## Description

`packages/catalyst-core/src/native/bridge/WebBridge.js` (JS-side dispatch layer that receives native callbacks) is not wired into the error registry shipped in #359. Found during a follow-up audit: dispatch failures (invalid interface, handler-not-registered, callback exceptions) still log via raw `console.error`/`console.warn` instead of going through `createError`/`CatalystError`.

This is a natural extension of #359 — `native/bridge/errors.js` (which handles native-error translation) already imports the shared registry; `WebBridge.js` sits right next to it and handles the other half of the same bridge (JS-side dispatch failures), but was left uncovered.

## Acceptance criteria

- [ ] WebBridge.js dispatch failures (invalid interface, unregistered handler, callback exceptions) throw/log via `createError` with appropriate `RUNTIME-NATIVE-xxx` codes (or new codes if the existing ones don't fit)
- [ ] No change to WebBridge's actual dispatch behavior — only the error reporting path

## Parent
Story: #343

mayankmahavar1mg:
## Summary

Wires `WebBridge.js` (JS-side bridge dispatch layer) into the error registry introduced in #359, per the proposal posted on #370.

Adds 5 new codes:
- `RUNTIME-NATIVE-018` — invalid callback interface
- `RUNTIME-NATIVE-019` — no handler registered for interface
- `RUNTIME-NATIVE-020` — registered handler threw
- `RUNTIME-NATIVE-021` — invalid `register()` call (non-function or invalid interface)
- `RUNTIME-NATIVE-022` — `WebBridge.init()` called outside a browser (genuine failure only)

## Scope

Only the pure-logging paths in `WebBridge.js` (`callback`, `register`, `unregister`'s valid case, `static init`) are wired in. The Promise-rejecting paths (`requestHapticFeedback`, `getDeviceInfoFromLegacyBridge`, `getDeviceInfoFromPluginBridge`, `getDeviceInfo`) are **not touched** — `shouldFallbackToLegacyDeviceInfo` pattern-matches specific error codes/messages from those rejections to decide whether to retry via the legacy bridge, and recoding them into catalyst codes risks breaking that fallback logic. This is flagged as an explicit, deliberate scope boundary, not an oversight.

`unregister()`'s "not registered, nothing to unregister" case is left as plain text — it's a benign no-op, not an error condition.

## Review finding fixed before commit

Initial version used the same code (`RUNTIME-NATIVE-022`) for both "init failed" and "already initialized" in `static init()`. The second case isn't a failure — the bridge is already up and working — so using the same code there produced a message that contradicted its own details ("could not be initialized" / "returning the existing instance"). Fixed: the already-initialized path stays a plain `console.warn`, no code; `RUNTIME-NATIVE-022` now means only the genuine no-`window` failure.

## Known limitation

`docUrl` links point at `blob/main/errors/...`. These stay dead not just until this PR merges, but until `story/6-error-handling` itself merges to `main` — a longer tail than a single-PR merge would imply, since this PR only targets the story branch.

## Test plan

- [x] Verified all 6 `ERROR_CODES.*` references in `WebBridge.js` resolve to real registry entries (no silent fallback to a generic code)
- [x] Loaded the actual (unmodified) `WebBridge.js` file in an isolated ESM harness with browser globals shimmed, and exercised: invalid callback interface, no handler registered, handler-threw (cause preserved), invalid register() with non-function, invalid register() with bad interface name, init() first-call vs already-initialized
- [x] Confirmed already-initialized `init()` no longer prints a code (fixed per above)
- [x] Regenerated `errors/**/*.md` + `errors/index.json`, 48 codes total (was 43 after #359)

Closes #370.

### DIFF
--- a/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
+++ b/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
@@ -1,0 +1,19 @@
+# RUNTIME-NATIVE-019
+
+**Category:** RUNTIME-NATIVE
+
+## Message
+
+No handler registered for this bridge interface
+
+## Details
+
+The native platform called back on an interface that has no JS-side handler registered yet.
+
+## Recoverable
+
+Yes — you can fix this and retry without restarting your workflow.
+
+## Suggested action
+
+Ensure WebBridge.register() is called for this interface before the native call arrives.
--- a/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
+++ b/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
@@ -8,7 +8,7 @@ No handler registered for this bridge interface
 
 ## Details
 
-The native platform called back on an interface that has no JS-side handler registered yet.
+Native sent a callback for this interface, but your webview hasn't registered a handler for it yet.
 
 ## Recoverable
 
@@ -16,4 +16,4 @@ Yes — you can fix this and retry without restarting your workflow.
 
 ## Suggested action
 
-Ensure WebBridge.register() is called for this interface before the native call arrives.
+Call WebBridge.register() for this interface before native calls arrive.
--- a/errors/index.json
+++ b/errors/index.json
@@ -354,9 +354,9 @@
     "RUNTIME-NATIVE-019": {
         "category": "RUNTIME-NATIVE",
         "message": "No handler registered for this bridge interface",
-        "details": "The native platform called back on an interface that has no JS-side handler registered yet.",
+        "details": "Native sent a callback for this interface, but your webview hasn't registered a handler for it yet.",
         "recoverable": true,
-        "suggestedAction": "Ensure WebBridge.register() is called for this interface before the native call arrives.",
+        "suggestedAction": "Call WebBridge.register() for this interface before native calls arrive.",
         "docUrl": "https://github.com/tata1mg/catalyst-core/blob/main/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md"
     },
     "RUNTIME-NATIVE-020": {
--- a/errors/index.json
+++ b/errors/index.json
@@ -342,5 +342,45 @@
         "recoverable": true,
         "suggestedAction": "Close other camera apps and try again",
         "docUrl": "https://github.com/tata1mg/catalyst-core/blob/main/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-017.md"
+    },
+    "RUNTIME-NATIVE-018": {
+        "category": "RUNTIME-NATIVE",
+        "message": "Invalid callback interface",
+        "details": "The native platform invoked WebBridge.callback() with an interface name that isn't recognized.",
+        "recoverable": false,
+        "suggestedAction": "Check the interface name against the registered NATIVE_CALLBACKS list; this usually indicates a native/JS version mismatch.",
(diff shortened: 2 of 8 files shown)


errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
```diff
@@ -0,0 +1,19 @@
...
+
+The native platform called back on an interface that has no JS-side handler registered yet.
+
+## Recoverable
+
+Yes — you can fix this and retry without restarting your workflow.
+
+## Suggested action
+
+Ensure WebBridge.register() is called for this interface before the native call arrives.
```
trinachoudhury1mg: is this error expected when native invoked a method on webbridge corresponding to which the mapped method in webview didnt exist as webbridge was not registered? 
If so, i do feel we need to visit the way we have described it. This is coming from somebody who knows the flow but not really the terminology used by catalyst team for it.

errors/index.json
```diff
@@ -342,5 +342,45 @@
         "recoverable": true,
         "suggestedAction": "Close other camera apps and try again",
         "docUrl": "https://github.com/tata1mg/catalyst-core/blob/main/errors/RUNTIME-NATIVE/RUNTIME-NATIVE-017.md"
+    },
+    "RUNTIME-NATIVE-018": {
+        "category": "RUNTIME-NATIVE",
+        "message": "Invalid callback interface",
+        "details": "The native platform invoked WebBridge.callback() with an interface name that isn't recognized.",
+        "recoverable": false,
```
trinachoudhury1mg, line 350: whats the definition of recoverable? Something that needs to be fixed at webview level and doesnt require a rebuild ?

trinachoudhury1mg (changes requested):

errors/RUNTIME-NATIVE/RUNTIME-NATIVE-019.md
mayankmahavar1mg: yeah that's exactly it — checked WebBridge.js, this fires when native calls back on an interface with no `register()` call yet on the JS side. agreed the wording is too internal, will reword to something like: "native sent a callback for `<interface>`, but no handler is registered for it yet. call `WebBridge.register()` before native calls arrive."

errors/index.json
mayankmahavar1mg: good catch — turns out `recoverable` was never actually defined anywhere, it was just true/false per code with no real rule behind it. rather than invent one now, we're dropping the field entirely. will clean it up across registry.js, index.js, and the docs generator.

mayankmahavar1mg:
reworded RUNTIME-NATIVE-019 to dev-facing language in 08f63a41. the `Recoverable` section will disappear from the .md once this rebases onto story/6-error-handling — we're dropping that field there (see #381).

mayankmahavar1mg:
Merging this now per Mayank's direction — the review has been silent since Aug 11 (formal `CHANGES_REQUESTED`), with no reply to our follow-up on Aug 17 either.

For the record, both requested changes landed:
- `RUNTIME-NATIVE-019` reworded to dev-facing language: `08f63a41`
- undefined `recoverable` field dropped repo-wide (including downstream categories added after this PR, and the CCA mirror): `18530a89`, `689a0bd5`, `d3f53124`

@trinachoudhury1mg if there's anything outstanding once you're back, happy to follow up in a new issue/PR.

PR #377, opened 2026-08-04
feat(errors): add AI-xxx error category for packages/catalyst-ai

### LINKED ISSUE #371: Add AI-xxx error category for packages/catalyst-ai
## Description

`packages/catalyst-ai` (multi-provider AI integration, `useAI`/`useCloudAI`/`useWebAI`/`useNativeAI` hooks + `route.js` server proxy) has zero error-code coverage today. Found during a follow-up audit after #359: every failure path is a raw `throw new Error(string)` or `console.error`, including:
- Provider API failures (auth/rate-limit/network) in `route.js` (multiple call sites, OpenAI/Gemini)
- Streaming/generate errors swallowed after logging in `route.js`
- Native-AI-bridge HTTP failures and several silent `catch (_) {}` blocks in `useNativeAI.js`
- Worker errors logged but not propagated in `useWebAI.js`

This needs its own category prefix (e.g. `AI-xxx`) added to the shared registry (`packages/catalyst-core/src/errors/registry.js`), following the pattern established in #359. Adding a new category is a precedent-setting decision — worth flagging in the design proposal for this issue, not just doing silently.

## Acceptance criteria

- [ ] New `AI-xxx` (or similar) category added to the shared registry
- [ ] Provider API failures, streaming errors, and native-AI-bridge failures in catalyst-ai use the new codes
- [ ] Silent `catch (_) {}` blocks in useNativeAI.js either surface a code or have documented justification for staying silent
- [ ] Docs generated for new codes via the existing `generateDocs.js` pattern

## Parent
Story: #343

mayankmahavar1mg:
Closes #371.

## Summary

Adds `AI-xxx` error codes for `packages/catalyst-ai`, per the design proposed and amended on #371:
- `AI-000` — foreign-wrapper for OpenAI/Gemini non-2xx responses (`wrapProviderError`, same pattern as `wrapForeignError` for build toolchains)
- `AI-001`..`AI-009` — catalyst-ai-owned codes for disabled/not-configured/invalid-body (route.js), native bridge/HTTP failures (`useNativeAI.js`), and worker failures (`useWebAI.js`)

Full rationale and 3 divergences from the original proposal are documented on #371.

## Notable fixes bundled in (found while implementing, same class as bugs caught on #359/#370)

- `useNativeAI.js`: `if (data.error) throw new Error(data.error)` inside the streaming SSE loop was nested inside a `catch (_) {}` meant only for `JSON.parse` failures on malformed chunks — the intentional throw was being swallowed and never reached `setError`. Same bug existed in the non-streaming path (masked there since it wasn't nested in a try/catch, but grouped fix for consistency).
- `errors/index.js` imported `getDocUrl` from `registry.js` but never re-exported it.

## Cross-package dependency note (flagging for review, touches Story 4 / API-freeze surface)

This adds a new export, `catalyst-core/errors` (resolves to `dist/errors/index.js`, built by the existing `prepare` script — no new build step). `catalyst-ai` depends on `catalyst-core: ">=0.3.0-0"`, and `route.js` — including the previously-standalone `/providers` endpoint — now resolves this export on every request path.

`packages/catalyst-ai/scripts/check-peer.js` currently only checks that `catalyst-core` resolves at all, not that it's new enough to expose `./errors`. Against an older catalyst-core this would fail with `ERR_PACKAGE_PATH_NOT_EXPORTED` inside the route handler's own catch block, producing a rejection rather than a clean response. Not fixed in this PR (version-gating `check-peer.js` felt like a separate decision) — flagging for maintainer input on whether it should block merge.

## Stack

```
feature/371-catalyst-ai   (this PR)
        ↓
feature/370-webbridge     (#373, → story/6-error-handling, awaiting review)
        ↓
story/6-error-handling
        ↓
main
```

Targets `feature/370-webbridge` since it depends on #370's registry additions, which haven't landed on the story branch yet.

### DIFF
--- a/packages/catalyst-ai/src/route.js
+++ b/packages/catalyst-ai/src/route.js
@@ -2,6 +2,14 @@ const express = require("express")
 
 const router = express.Router()
 
+// catalyst-core/errors is ESM-only; route.js stays CJS (loaded via require.resolve
+// in expressServer.js), so we load it lazily via dynamic import and cache the module.
+let errorsModulePromise = null
+function loadErrors() {
+    if (!errorsModulePromise) errorsModulePromise = import("catalyst-core/errors")
+    return errorsModulePromise
+}
+
 // Hard ceiling on how long a single /stream request may run upstream — protects
 // against a hung provider connection holding a request open indefinitely.
 const STREAM_TIMEOUT_MS = 120_000
@@ -50,6 +58,13 @@ function sseHeaders(res) {
     res.setHeader("Connection", "keep-alive")
     res.setHeader("X-Accel-Buffering", "no")
     res.flushHeaders()
+}
+
+// Wraps a non-ok provider response as a CatalystError (AI-000) and throws it,
+// preserving the upstream status/body verbatim as `cause`.
+async function throwProviderError(provider, response) {
+    const { wrapProviderError } = await loadErrors()
+    throw wrapProviderError(provider, response.status, await response.text())
 }
 
 // ─── usage normalization ──────────────────────────────────────────────────────
@@ -134,7 +149,7 @@ async function openaiStream({ apiKey, model, messages, genConfig, conversationId
             body: JSON.stringify(body),
             signal,
         })
-        if (!response.ok) throw new Error(`OpenAI Responses API error (${response.status}): ${await response.text()}`)
+        if (!response.ok) await throwProviderError("openai", response)
 
         const reader = response.body.getReader()
         const decoder = new TextDecoder()
@@ -187,7 +202,7 @@ async function openaiStream({ apiKey, model, messages, genConfig, conversationId
             body: JSON.stringify(body),
             signal,
         })
-        if (!response.ok) throw new Error(`OpenAI API error (${response.status}): ${await response.text()}`)
+        if (!response.ok) await throwProviderError("openai", response)
 
         const reader = response.body.getReader()
         const decoder = new TextDecoder()
@@ -238,7 +253,7 @@ async function openaiGenerate({ apiKey, model, messages, genConfig, conversation
             headers: { "Content-Type": "application/json", Authorization: `Bearer ${apiKey}` },
             body: JSON.stringify(body),
         })
-        if (!response.ok) throw new Error(`OpenAI Responses API error (${response.status}): ${await response.text()}`)
+        if (!response.ok) await throwProviderError("openai", response)
         const data = await response.json()
         const output = data.output?.find((o) => o.type === "message")?.content?.find((c) => c.type === "output_text")?.text ?? ""
         return { output, conversationId: data.id ?? null, usage: normalizeOpenAIResponsesUsage(data.usage, model) }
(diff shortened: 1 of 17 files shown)


packages/catalyst-ai/src/route.js
```diff
@@ -2,6 +2,14 @@ const express = require("express")
 
 const router = express.Router()
 
+// catalyst-core/errors is ESM-only; route.js stays CJS (loaded via require.resolve
```
trinachoudhury1mg, line 5: why is route.js cjs?

packages/catalyst-core/package.json
```diff
@@ -24,6 +24,7 @@
         "./document": "./dist/server/renderer/document/index.jsx",
         "./WebBridge": "./dist/native/bridge/WebBridge.js",
         "./hooks": "./dist/native/bridge/hooks.js",
+        "./errors": "./dist/errors/index.js",
```
trinachoudhury1mg, line 27: by exposing this outside are we recommending folks to use CatalystError instead of Error ?

packages/catalyst-ai/src/useWebAI.js
```diff
@@ -244,7 +245,10 @@ export function useWebAI({
             try {
                 worker = new Worker(getWorkerBlobUrl(), { type: "module" })
             } catch (err) {
-                setError(new Error("[catalyst-ai/useWebAI] module worker unavailable: " + (err.message || String(err))))
+                setError(createError(ERROR_CODES.AI_WEB_WORKER_UNAVAILABLE, {
+                    message: "Constructing the module Worker failed: " + (err.message || String(err)),
```
trinachoudhury1mg: do we need || String(err)  when we are adding cause as well ?

packages/catalyst-ai/src/route.js
mayankmahavar1mg: it's loaded via require.resolve in expressServer.js, so it has to stay CJS. catalyst-core/errors is ESM-only though, so we load it lazily with a cached dynamic import() instead — left a comment above loadErrors() explaining it.

packages/catalyst-core/package.json
mayankmahavar1mg: no, not meant for app-level use — this export is for our own packages (catalyst-ai here) to share the registry, not something we want app devs importing directly. will make that clearer, maybe in a comment on the export itself.

packages/catalyst-ai/src/useWebAI.js
mayankmahavar1mg: good catch, no we don't — formatError already does `cause.message || String(cause)` when rendering, so this was just duplicating that. will drop the manual concat.

mayankmahavar1mg:
pushed in 9487887b — dropped the redundant message concat in useWebAI.js (formatError already renders it from `cause`), and added a comment on errors/index.js clarifying `catalyst-core/errors` is for our own packages, not meant for app-level use.

PR #379, opened 2026-08-06
feat: sub-path base + HMR config for reverse-proxy dev setups

yadinesh1:
## Problem

Apps served under a sub-path (e.g. `/o-mweb`) behind an SSL reverse proxy (nginx/caddy) can't use catalyst-core's dev server as-is. Vite's `base` defaults to `"/"`, so every dev URL the framework emits lacks the prefix and a path-routed proxy sends it to a different upstream — the client entry, HMR socket, and the react-refresh preamble all load from the wrong origin. Symptoms cascade:

1. **White screen / 404s** — `/@vite/client`, `/src/js/…` and the SSR-injected `/client/index.js` have no base prefix, so under a sub-path proxy they hit a different service and React never hydrates.
2. **`$RefreshSig$ is not defined`** — the react-refresh preamble hardcoded `import … from "/@react-refresh"`; that 404s under a sub-path, so `window.$RefreshReg$/$RefreshSig$` are never installed and every `@vitejs/plugin-react` module throws.
3. **No HMR config for SSL proxies** — no escape hatch in `buildConfig.js` for the HMR `clientPort`/`path` when the browser connects through an HTTPS proxy on 443.

## Solution

`buildConfig.devServer` is the single place to configure the dev server, with `devServer.base` as the sole source of truth for the mount path (default `"/"`).

### `vite.config.js` + new `resolveDevServerConfig.js`
- `sharedViteConfig` (plain object) holds everything common to dev + prod; the `default` export is an **async factory** that reads `buildConfig.devServer`.
- Base / HMR / fs / passthrough resolution is extracted into **pure, unit-tested helpers**.
- `server.hmr` merges `devServer.hmr` (`{ clientPort: 443, path: "__vite_hmr" }` for proxy setups); other `devServer` keys (`proxy`, `cors`, `allowedHosts`, …) are forwarded verbatim.
- `fs.allow` always includes the framework paths and can only be **extended** by `devServer.fs.allow` — apps can't replace it or relax `strict`, so buildConfig can't widen dev-server filesystem exposure.

### `vite.config.server.js` / `vite.config.client.js`
- Both spread the named `sharedViteConfig` object, so the production SSR and client builds keep `cacheDir`, resolve aliases, css, `assetsInclude` and base plugins.

### `server/expressServer.js` + `Body.jsx` / `FastRefresh.jsx`
- After the dev server starts, the resolved `vite.config.base` is exposed to the SSR renderer, which base-prefixes the dev client entry and the react-refresh preamble. Because the renderer reads Vite's own resolved base, the injected URLs can never diverge from the path Vite serves under.

## Usage

```js
// buildConfig.js  (in any catalyst app)
export default {
  devServer: {
    base: "/o-mweb",                               // single source of truth; default "/"
    hmr: { clientPort: 443, path: "__vite_hmr" },  // only for SSL proxy setups
    allowedHosts: ["local.example.com"],
  },
}
```

## Review feedback addressed

| # | Severity | Fix |
| --- | --- | --- |
| 1 | 🔴 Critical | Both build configs spread the plain `sharedViteConfig` object; the SSR build no longer spreads the async factory function (which silently dropped the whole shared config from the prod SSR bundle). |
| 2 | 🟠 Warning | fs allowlist hardened — passthrough `serverOverrides` are spread **first** so framework-controlled `hmr`/`fs` always win; `fs.allow` can only be extended, never replaced or have `strict` relaxed. Logic moved to a pure, tested helper. |
| 3 | 🟠 Warning | The renderer no longer reads a possibly-divergent env var; it uses Vite's resolved `base` (from `devServer.base`), trailing-slash-normalized before URL joins. |
| 4 | 🔵 Info | Added `test/vite/resolveDevServerConfig.test.js` (Node's built-in test runner, zero new deps) covering base precedence/normalization, the fs.allow merge, and hmr/passthrough handling. `npm run test:unit`. |

## Backward compatibility

- Apps without `buildConfig.devServer.base` get `base: "/"` — identical to current behaviour.
- `sharedViteConfig` is spread the same way the old default export was, in both build configs.
- Root-mounted apps get a `""` prefix → `/client/index.js` and `/@react-refresh` exactly as before.

### DIFF
--- a/packages/catalyst-core/src/server/renderer/document/Body.jsx
+++ b/packages/catalyst-core/src/server/renderer/document/Body.jsx
@@ -30,7 +30,10 @@ export function Body(props) {
                     __html: `window.__SAFE_AREA_INITIAL__ = ${JSON.stringify(safeArea)}; window.__CATALYST_NATIVE_WEBVIEW__ = ${nativeWebView ? "true" : "false"}`,
                 }}
             />
-            {process.env.NODE_ENV === "development" && <script type="module" src="/client/index.js"></script>}
+            {process.env.NODE_ENV === "development" && (
+                // Base pinned to Vite's resolved base by server/expressServer.js.
+                <script type="module" src={`${process.env.APP_MOUNT_PATH || ""}/client/index.js`}></script>
+            )}
             {jsx}
             <script
                 /* eslint-disable */
--- a/packages/catalyst-core/package.json
+++ b/packages/catalyst-core/package.json
@@ -30,6 +30,7 @@
     "scripts": {
         "lint": "eslint .",
         "lint-staged": "lint-staged",
+        "test:unit": "node --test \"test/**/*.test.js\"",
         "prettify": "prettier . --write",
         "prepare": "node ../../scripts/safe-rimraf.js ./dist && cp -r ./src ./dist && babel src/native --out-dir dist/native --config-file ./babel.native.config.cjs --ignore '**/.build/**' && babel src/sentry.js --out-file dist/sentry.cjs --config-file ./babel.native.config.cjs",
         "prepublishOnly": "npm run prepare",
--- a/packages/catalyst-core/src/server/expressServer.js
+++ b/packages/catalyst-core/src/server/expressServer.js
@@ -10,6 +10,7 @@ import pc from "picocolors"
 import fs from "fs"
 const { cyan, yellow, green } = pc
 
+import { toMountPathPrefix } from "../vite/resolveDevServerConfig.js"
 import { validateMiddleware, safeCall } from "./utils/validator.js"
 import { botDetectionMiddleware } from "./utils/botDetectionMiddleware.js"
 import { cjsRequire } from "./utils/cjsRequire.js"
@@ -208,6 +209,12 @@ async function createServer() {
             appType: "custom",
             root: process.env.src_path,
         })
+
+        // Expose Vite's resolved base to the SSR renderer so the URLs it injects
+        // (client entry, react-refresh preamble) match the path Vite serves under.
+        // Vite normalizes base with a trailing slash; trim it for clean URL joins.
+        process.env.APP_MOUNT_PATH = toMountPathPrefix(vite.config.base)
+
         app.use(vite.middlewares)
     }
 
--- a/packages/catalyst-core/src/vite/FastRefresh.jsx
+++ b/packages/catalyst-core/src/vite/FastRefresh.jsx
@@ -6,8 +6,10 @@ const FastRefresh = () => {
             type="module"
             // eslint-disable-next-line risxss/catch-potential-xss-react
             dangerouslySetInnerHTML={{
+                // Base-prefixed so the preamble resolves under a sub-path; otherwise
+                // plugin-react modules throw "$RefreshSig$ is not defined".
                 __html: `
-        import { injectIntoGlobalHook } from "/@react-refresh";
+        import { injectIntoGlobalHook } from "${process.env.APP_MOUNT_PATH || ""}/@react-refresh";
         injectIntoGlobalHook(window);
         window.$RefreshReg$ = () => {};
         window.$RefreshSig$ = () => (type) => type;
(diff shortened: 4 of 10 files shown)


mayankmahavar1mg:
## Review Summary

**Overall Rating: ⭐⭐☆☆☆ (2/5)**

Good intent and clean documentation, but this ships a silent production-breaking regression plus a security-relevant allowlist bypass. Would be 4★ once the two blocking issues below are fixed.

| # | Severity | File | Line(s) | Issue | Rating |
|---|----------|------|---------|-------|--------|
| 1 | 🔴 Critical | `packages/catalyst-core/src/vite/vite.config.server.js` | 4, 17, 23, 28 | `baseConfig` default import from `vite.config.js` is now an async factory function (not an object), but this file still does `...baseConfig`, `baseConfig.plugins`, `baseConfig.build` as if it were the old plain object. The production SSR build silently loses `cacheDir`, resolve aliases, CSS config, `assetsInclude`, and base plugins. Won't surface in this repo's own dev/CI loop — only after `npm run build` ships and a consumer runs SSR in production. `vite.config.client.js` was correctly updated to use the new `sharedViteConfig` export; this file was missed. | ⭐☆☆☆☆ |
| 2 | 🟠 Warning | `packages/catalyst-core/src/vite/vite.config.js` | ~515–530 | `...serverOverrides` (remaining keys of `buildConfig.devServer` after destructuring out `base`/`hmr`) is spread *after* the hardcoded `fs: { allow: [...] }` block, so any app's `buildConfig.js` can override it — e.g. `devServer.fs.allow: ['/']` or `fs.strict: false` — widening dev-server filesystem exposure. Notable given this PR's own use case of proxying the dev server to a public hostname via `allowedHosts`. | ⭐⭐☆☆☆ |
| 3 | 🟠 Warning | `packages/catalyst-core/src/server/renderer/document/Body.jsx` | 33–38 | `` `${process.env.APP_MOUNT_PATH || ""}/client/index.js` `` isn't normalized. `APP_MOUNT_PATH="/"` produces `//client/index.js` (protocol-relative URL, breaks HMR); any trailing-slash value produces a double slash. | ⭐⭐☆☆☆ |
| 4 | 🔵 Info | `packages/catalyst-core/src/vite/loadCustomViteConfig.js` | new `devServer` block | No tests added for the new `devServer` precedence (`base`/`hmr`/`allowedHosts`), the `fs` merge behavior, or the `Body.jsx` mount-path concatenation — despite touching production build config resolution and a security-relevant allowlist. | ⭐⭐⭐☆☆ |
| 5 | 🟢 Positive | `loadCustomViteConfig.js` | JSDoc | Clear documentation of the `devServer` option with a concrete reverse-proxy example. | ⭐⭐⭐⭐⭐ |
| 6 | 🟢 Positive | `vite.config.js` | `sharedViteConfig` split | Reasonable factoring of static vs. dev-only config — the approach is sound, just not fully propagated to `vite.config.server.js`. | ⭐⭐⭐⭐☆ |

**Blocking:** #1 and #2 should be fixed before merge. #3 is worth fixing in the same pass. #4 is reasonable to request given the blast radius of #1/#2.

packages/catalyst-core/src/server/renderer/document/Body.jsx
```diff
@@ -30,7 +30,12 @@ export function Body(props) {
                     __html: `window.__SAFE_AREA_INITIAL__ = ${JSON.stringify(safeArea)}; window.__CATALYST_NATIVE_WEBVIEW__ = ${nativeWebView ? "true" : "false"}`,
                 }}
             />
-            {process.env.NODE_ENV === "development" && <script type="module" src="/client/index.js"></script>}
+            {process.env.NODE_ENV === "development" && (
+                <script
+                    type="module"
+                    src={`${process.env.APP_MOUNT_PATH || ""}/client/index.js`}
```
Uttkarsh-Srivastava: pass this through buildConfig

Uttkarsh-Srivastava (changes requested):

Issue #372, opened 2026-08-03
Add error category for packages/create-catalyst-app (CCA)


mayankmahavar1mg:
## Description

`packages/create-catalyst-app` (the CLI that scaffolds new catalyst-core projects, `scripts/cli.cjs`) has zero error-code coverage today. Found during a follow-up audit after #359: invalid project name, directory conflicts, MCP setup failures, and bad CLI option values are all raw `console.error`/`throw new Error(string)` + `process.exit(1)`, with no structured codes.

This needs its own category prefix (e.g. `CCA-xxx` or `SCAFFOLD-xxx`) added to the shared registry, following the #359 pattern. Since CCA is a separate package from catalyst-core, confirm whether it should depend on `packages/catalyst-core/src/errors` directly or needs its own lightweight copy of the pattern (dependency direction between these packages should be decided as part of the design proposal for this issue).

## Acceptance criteria

- [ ] Decision made on how CCA references the shared error registry (direct dependency vs local copy) — documented in the issue before implementation
- [ ] New category added; invalid project name, directory conflicts, MCP setup failures, and bad option values use structured codes
- [ ] Docs generated for new codes via the existing `generateDocs.js` pattern

## Parent
Story: #343

mayankmahavar1mg:
## Design proposal

**Dependency direction: local copy, not a dependency on `catalyst-core`.**

Traced the actual execution flow in `scripts/cli.cjs`: the CLI packs and extracts **itself** (`create-catalyst-app`'s own package, via `packNpmPackage()`/`extractSubdirectory()`) to scaffold the new project directory, then runs `npm i` *inside that new project directory* — that's the point where `catalyst-core` actually gets installed, into `<projectName>/node_modules`, not into the CLI process's own environment. Every one of CCA's own error paths — invalid project name, directory-already-exists conflict, npm-pack failure, tar-extraction failure, and MCP setup (`runMcpSetup`) — runs before or during that step, when `catalyst-core` is not present or reachable from the CLI process. Adding `catalyst-core` as a real dependency of `create-catalyst-app` would mean every `npx create-catalyst-app` install pulls in the full framework package just to get error codes, which is heavy for the ecosystem's most common entry point.

[...]

Issue #329, opened 2026-08-03
Epic: Catalyst-Core 1.0.0 Stable Release


mayankmahavar1mg:
## Goal

Ship catalyst-core `1.0.0` — feature-complete and API-locked, backed by automated test coverage and CI enforcement — per the release model defined in #177 and the error-handling scope defined in #155.

## Gate

All sub-issues under Stories 1–6 must close before the `1.0.0-rc.1` cut. Story 7 (landing in-flight PRs) lands opportunistically. Story 8 (RFC #148) requires a maintainer decision before any implementation is scheduled — it is not yet gating.

## Stories

- [ ] Story 1 — Resolve the runtime-adapter decision (blocks API freeze)
- [ ] Story 2 — Fix correctness bugs blocking stable
- [ ] Story 3 — Test coverage and CI enforcement
- [ ] Story 4 — API freeze at RC cut
- [ ] Story 5 — Release-gating documentation
- [ ] Story 6 — Error handling (RFC #155, Phase 1)
- [ ] Story 7 — Land in-flight PRs
- [ ] Story 8 — RFC #148 decision

## References

- #177 — Catalyst Framework Release Flow
- #155 — Improved Error Handling for Catalyst-Core
- #148 — Version Management Strategy
- #313 / #315 / #316 — Runtime Adapter seam proposal

Issue #335, opened 2026-08-03
Make preServerInit failures stop server startup


mayankmahavar1mg:
## Description

Ref #298. A `preServerInit` lifecycle failure currently logs and continues instead of halting server startup, so a broken server can come up looking healthy.

## Acceptance criteria

- [ ] `preServerInit` failure stops server startup
- [ ] Clear error surfaced explaining why startup halted

## Parent
Story: #332

mayankmahavar1mg:
heads up — when this lands, safeCallNamed and safeCall in server/utils/validator.js should probably get merged into one (safeCallNamed is just safeCall + a hook name, only called once). holding off until this is fixed since it might remove safeCallNamed's only caller. see PR #381 review thread.

PR #381, opened 2026-08-06
Story 6: Error Handling (RFC #155)

### LINKED ISSUE #343: Story: Error handling (RFC #155, Phase 1)
## Goal

Implement Phase 1 of RFC #155 (error handling). Phase 2 items (FAQ, report-issue command, Sentry/Crashlytics) are explicitly out of scope for 1.0 per the maintainer-approved phased scope.

## Tasks

- [ ] Implement unified error codes
- [ ] Implement three output modes
- [ ] Build catalyst:doctor preflight validation command
- [ ] Publish versioned error documentation site
- [ ] Add interactive Flock terminal-link integration

## Parent

Epic: #329

mayankmahavar1mg:
## Summary
Story 6 (Error Handling, RFC #155) for Epic #329. Introduces the unified error-code registry (PREFLIGHT/PROCESS/BUNDLE/IOS/ANDROID/RUNTIME-NATIVE) and wires it through catalyst-core's build/native surfaces.

Stacked as: `feature/371-catalyst-ai` (#377) → `feature/370-webbridge` (#373) → `story/6-error-handling` (this PR) → `epic/329` → `main`.

Task PRs #373 and #377 must merge into this branch before this PR is ready to merge into `epic/329`.

## Test plan
- [ ] #373 merged
- [ ] #377 merged
- [ ] #372 (create-catalyst-app error category) merged


packages/catalyst-core/src/errors/registry.js
```diff
@@ -0,0 +1,378 @@
...
+}
+
+export function getDefinition(code) {
+    return ERROR_DEFINITIONS[code] || ERROR_DEFINITIONS[ERROR_CODES.RUNTIME_NATIVE_INTERNAL_ERROR]
+}
+
+export function getDocUrl(code) {
+    const def = ERROR_DEFINITIONS[code]
+    if (!def) return null
+    return docUrl(def.category, code)
```
trinachoudhury1mg: should directly calculate 
 return `${REPO_BLOB_BASE}/${category}/${code}.md` 
here instead of a another un-exported function

packages/catalyst-core/src/errors/index.js
```diff
@@ -0,0 +1,91 @@
...
+}
+
+const STAGE_WRAPPER_CODE = {
+    BUNDLE: ERROR_CODES.BUNDLE_UPSTREAM_ERROR,
+    IOS: ERROR_CODES.IOS_UPSTREAM_ERROR,
+    ANDROID: ERROR_CODES.ANDROID_UPSTREAM_ERROR,
+}
+
+const STAGE_UPSTREAM_NAME = {
+    BUNDLE: "Vite/Rollup",
```
trinachoudhury1mg, line 50: Vite/Rollup this string gets repeated across multiple files. how will u ensure we dont miss updating it at all places.
Assuming same for other upstream names as well

packages/catalyst-core/src/errors/index.js
```diff
@@ -0,0 +1,91 @@
...
+        cause: err,
+    })
+    return wrapped
+}
+
+/**
+ * Format a CatalystError for terminal output. Foreign-wrapped errors always
+ * show the original upstream message/code, never a paraphrase of it.
+ */
+export function formatError(err) {
```
trinachoudhury1mg: why not invoke formatError from wrapForeignError or wrapError methods directly and return formatted error?

packages/catalyst-core/src/errors/index.js
```diff
@@ -0,0 +1,91 @@
...
+    const lines = [`[${err.code}] ${err.message}`]
+    if (err.cause) {
+        const causeMessage = err.cause.message || String(err.cause)
+        lines.push(`→ ${causeMessage}`)
```
trinachoudhury1mg, line 123: why no title "Cause:" or "Description :" before cause and details ?
and  is cause and details mutually exclusive ?

packages/catalyst-core/src/errors/registry.js
```diff
@@ -0,0 +1,378 @@
...
+    RUNTIME_NATIVE_INTERNAL_ERROR: "RUNTIME-NATIVE-015",
+    RUNTIME_NATIVE_CAMERA_UNAVAILABLE: "RUNTIME-NATIVE-016",
+    RUNTIME_NATIVE_CAMERA_IN_USE: "RUNTIME-NATIVE-017",
+}
+
+// One entry per catalyst-owned code. Foreign/wrapped errors (see wrapForeignError)
+// intentionally do not get an entry here beyond their per-stage wrapper code —
+// we don't author docs or specific meanings for errors we don't originate.
+export const ERROR_DEFINITIONS = {
+    [ERROR_CODES.PREFLIGHT_CONFIG_MISSING]: {
```
trinachoudhury1mg, line 78: whats the reason to have duplicity of error codes here and in errors/index.json ?

packages/catalyst-core/src/errors/registry.js
```diff
@@ -0,0 +1,378 @@
...
+    [ERROR_CODES.PREFLIGHT_MIDDLEWARE_NOT_FUNCTION]: {
+        category: "PREFLIGHT",
+        defaultMessage: "addMiddlewares should be a function present in server/server.js",
+        defaultDetails: "addMiddlewares was found but is not a function.",
+        recoverable: true,
+        suggestedAction: "Ensure addMiddlewares is exported as a function.",
+    },
+    [ERROR_CODES.PREFLIGHT_REDUCER_MISSING]: {
+        category: "PREFLIGHT",
+        defaultMessage: "reducer not found in src/js/containers/App/reducer",
```
trinachoudhury1mg, line 159: whats the diff between message and details?

packages/catalyst-core/src/errors/registry.js
```diff
@@ -0,0 +1,378 @@
...
+        defaultDetails: "This feature is not supported on the current platform",
+        recoverable: false,
+        suggestedAction: "Try using an alternative method or device",
+    },
+    [ERROR_CODES.RUNTIME_NATIVE_INTERNAL_ERROR]: {
+        category: "RUNTIME-NATIVE",
+        defaultMessage: "An unexpected error occurred",
+        defaultDetails: "An internal error occurred while processing the request",
+        recoverable: true,
+        suggestedAction: "Try again or contact support if the problem persists",
```
trinachoudhury1mg, line 329: shouldnt we ask the user to leverage mcp and share the error through a command already part of mcp - which will pick log files as the error mapping failed the error i assume was still logged?

packages/catalyst-core/src/server/utils/validator.js
```diff
@@ -138,15 +133,35 @@ const validateCustomDocument = (fn) => {
...
+        console.error(formatError(wrapped))
+    }
+}
+
+/**
+ * Same as safeCall, but names the specific hook (e.g. "preServerInit") in the
+ * error so it's identifiable rather than generic. Used at call sites that know
+ * which hook they're invoking.
+ */
+const safeCallNamed = (hookName, fn, ...args) => {
```
trinachoudhury1mg: i think we could have clubbed safeCall and this in one. 
Not many occurrences of safeCall exist in the codebase

packages/catalyst-core/src/scripts/validator.js
```diff
@@ -17,9 +17,8 @@ const validatePreInitServer = (fn) => {
 
 const validateMiddleware = (fn) => {
     try {
-        if (!fn) throw new Error("addMiddlewares named function not found in server/server.js")
-        if (typeof fn !== "function")
-            throw new Error("addMiddlewares should be function present in server/server.js")
+        if (!fn) throw createError(ERROR_CODES.PREFLIGHT_MIDDLEWARE_MISSING)
```
trinachoudhury1mg, line 20: we dont require formatting here?

packages/catalyst-core/src/errors/registry.js
mayankmahavar1mg: fair, `docUrl()` only has this one caller. will inline it and drop the helper.

packages/catalyst-core/src/errors/index.js
mayankmahavar1mg: it's actually not repeated — `STAGE_UPSTREAM_NAME` is the one place it's defined, `wrapForeignError` is the only reader. same for Xcode/CocoaPods and Gradle. single source, no sync needed.

packages/catalyst-core/src/errors/index.js
mayankmahavar1mg: makes sense, every call site does wrap-then-format back to back anyway. one wrinkle: handler.jsx also logs the raw unformatted error separately in non-debug mode, so we'd need to keep that available too. thinking a version that takes outputMode/debugEnv and returns the formatted string, but still hands back the raw error alongside. open to how you'd want that shaped.

packages/catalyst-core/src/errors/index.js
mayankmahavar1mg: yep, mutually exclusive by design — if there's a cause we show that (it already has the real explanation), otherwise we fall back to details. no explicit label right now, just the arrow (`→`). happy to add labels if you think it reads better that way.

packages/catalyst-core/src/errors/registry.js
mayankmahavar1mg: not actually duplicated — index.json and all the per-code .md files are generated from registry.js by generateDocs.js. registry.js is the only place anyone hand-edits.

packages/catalyst-core/src/errors/registry.js
mayankmahavar1mg: message is the short one-liner shown as the headline (`[CODE] message`), details is the longer explanation shown in the body when there's no cause. so also tied to the mutual-exclusivity thing from the other thread.

packages/catalyst-core/src/errors/registry.js
mayankmahavar1mg: like the idea but feels like it belongs to the diagnostics-stream work (#395) rather than baked into this one fallback definition's suggestedAction. want to pick it up there instead of here?

packages/catalyst-core/src/server/utils/validator.js
mayankmahavar1mg: agreed, safeCallNamed is really just safeCall + a hook name, and it's only called once (preServerInit). holding off on merging them though — #335 is about making preServerInit failures actually halt startup instead of being swallowed here, which might remove safeCallNamed's only caller. want to merge them after #335 lands so we're not redoing it.

packages/catalyst-core/src/scripts/validator.js
mayankmahavar1mg: intentional — these throws get caught right there in the same try/catch and routed to handleError(), which is the one place formatError runs. the throw site just needs the structured error, not a rendered string.

mayankmahavar1mg:
pushed fixes from the review threads — dropped `recoverable` (was never actually defined, just true/false per code) and inlined `docUrl()`, in 18530a89. errors/*.md + index.json regenerated to match.

mayankmahavar1mg:
## CodeRabbit Review

Ran `coderabbit review --agent --base epic/329` against this PR's diff (23 commits). **16 findings** — 0 critical, 4 major, 12 minor.

### Major
1. **`packages/catalyst-core/src/server/utils/validator.js`** — `safeCall`/`safeCallNamed` don't route rejected async-hook promises through the structured error path, only sync throws are caught today.
2. **`packages/catalyst-core/mcp_v2/tools/errors.js`** — `explain_error`'s index path points at the repo-level `errors/` dir, which won't resolve once catalyst-core is installed as a real npm dependency outside this monorepo.
3. **`packages/catalyst-ai/src/route.js`** — AI-000 server-side logs drop the upstream cause message, hurting debuggability (client response correctly still withholds it).
4. **`skills/new-feature/SKILL.md`** (2 findings) — worktree/sub-issue creation steps assume prior session context rather than being self-contained (don't fetch/branch from `origin/main` explicitly; `addSubIssue` GraphQL isn't documented inline).

### Notable minor
- **Real (low-severity) bug**: `errors.js`'s `explain_error` uses `_index[code]` instead of an own-property check — a code string like `"toString"` gets misreported as catalyst-owned.
- `buildErrorFormat.js` assumes thrown build errors are always `Error` instances (breaks on thrown strings/null).
- CCA-009 in `create-catalyst-app/scripts/cli.cjs` ignores `--verbose`/`--debug` flags.
- ~8 findings are wording/precision nits in the generated `errors/*.md` docs (mismatched file paths, ambiguous named-vs-default-export phrasing) — cosmetic, not functional.

None of these are blocking/security-severity. Working through the 4 majors + the real minor bug now; will push fixes as follow-up commits.

mayankmahavar1mg:
Pushed fixes for 6 of the 16 CodeRabbit findings in `fb0aec68`:

- **safeCall/safeCallNamed** now async — rejected promises from async hooks route through the structured-error path instead of becoming unhandled rejections. Verified sync hooks are unaffected (awaiting a non-Promise return still resolves synchronously, so `res.headersSent` checks in handler.jsx keep working).
- **explain_error** package-install path fixed — added `mcp_v2/scripts/copyErrorsIndex.cjs`, wired into `prepare`/`prepublishOnly`, so `dist/errors-index.json` ships inside the package instead of relying on the (unpublished) repo-root `errors/` dir.
- **explain_error** own-property lookup bug fixed — `"toString"`/`"constructor"` no longer get misreported as catalyst-owned codes.
- **catalyst-ai route.js** — AI-000 stream/generate logs now include the upstream cause (bounded to 500 chars) server-side; client response unchanged.
- **CCA-009** in `create-catalyst-app/cli.cjs` now respects `--verbose`/`--debug` via the file's existing `printError()` helper.
- **buildErrorFormat.js** normalizes non-Error thrown values before touching `.message`/`.stack`.

Remaining findings are doc-text wording nits in the generated `errors/*.md` content (8 of them, cosmetic — file-path/wording precision, not functional) and two `skills/new-feature/SKILL.md` self-containedness notes. Leaving those for a follow-up pass rather than blocking this PR further.

mayankmahavar1mg:
## CodeRabbit Re-review

Ran `coderabbit review --agent --base epic/329` again after the fix commit (`fb0aec68`).

**0 findings.** Clean pass across all 100 files in the diff.

Issue #360, opened 2026-08-03
Implement three output modes (default/verbose/debug)


mayankmahavar1mg:
## Description
Implement the three output modes defined in RFC #155: default (minimal, scannable, with solution), --verbose (detailed context + environment info), --debug (complete diagnostics + stack traces).

## Acceptance criteria
- [x] All three modes implemented and selectable via CLI flag
- [x] Output format matches RFC #155 spec for each mode

Implemented across all three module systems in the codebase (catalyst-core ESM, create-catalyst-app CJS, src/native CJS) per the posted design proposal and amendment comment (see below).

## Parent
Story: #343

## PR
#393 (stacked: #391 -> #390 -> #393)

## Related work found along the way
- #389: picocolors -> ansis migration (prerequisite — picocolors can't correctly color multi-line strings)
- #392: pre-existing ERR_REQUIRE_ESM bug found during example-app verification, filed separately (not caused by this work)

mayankmahavar1mg:
## Design proposal: three output modes (#360)

Scope: **catalyst-core (ESM)**, **create-catalyst-app/CCA (CJS)**, and **src/native buildApp scripts (CJS)** — all three, per discussion. Three separate implementations of the same mode semantics, because they're three different module systems that can't share code (see below).

### The constraint: 3 module systems, one of which can't even reach `formatError` today

| Surface | Module system | Current state |
|---|---|---|
| `errors/index.js` (`formatError`) | ESM | shared by `build.js`, `server/utils/validator.js`, `handler.jsx` (server, post-#386), `WebBridge.js` (**browser** — no `process.argv`, no TTY) |
| `create-catalyst-app/scripts/errors.cjs` | CJS | deliberate standalone duplicate (CCA packs/extracts itself before catalyst-core is installed — can't depend on it) |
| `src/native/buildAppAndroid.js`, `buildAppIos.js` | CJS | **not using `formatError` at all** — hand-written inline string literals (`` `[ANDROID-000] Build failed...` ``), because `src/native` is CJS-only and can't synchronously `require()` the ESM `errors/index.js` under Node 20 (existing comment in both files explains this) |

So this is three implementations of the same mode logic, not one shared function called from three places.

### Mode resolution: CLI flag, forwarded to child processes via env var

[...]

PR #390, opened 2026-08-06
chore: migrate picocolors to ansis

### LINKED ISSUE #389: Migrate picocolors to ansis (needed for #360's multi-line boxed output)
## Description

Migrate from `picocolors` to `ansis` across catalyst-core and create-catalyst-app.

Found while designing #360 (output modes): the new `--verbose`/`--debug` formats are multi-line, boxed (`┌─│└` borders), and colored. `picocolors` does not re-apply color codes across `\n` line breaks within a single colored multi-line string — colors can leak or break across lines. `ansis` correctly inserts a style break at each newline, and is comparably fast to `picocolors` in realistic (multi-style) benchmarks, unlike `chalk` which is meaningfully slower in all cases.

Sources: [picocolors GitHub](https://github.com/alexeyraspopov/picocolors), [ansis GitHub](https://github.com/webdiscus/ansis), [chalk vs picocolors – npm trends](https://npmtrends.com/chalk-vs-picocolors)

Doing this as a standalone migration first, rather than folding it into #360, since it touches 14 files across two packages — unrelated in scope to RFC #155's output-mode work, and mixing the two would make #360's diff harder to review.

## Files using `picocolors` today

catalyst-core:
- `src/errors/index.js`
- `src/logger.js`
- `src/native/terminalProgress.js`
- `src/scripts/plugins.js`
- `src/scripts/scriptUtils.js`
- `src/scripts/validator.js` (dead code — confirmed only referenced by a commented-out import in `loadEnvironmentVariables.js`; migrate for consistency but low risk)
- `src/server/expressServer.js`
- `src/server/utils/validator.js`

create-catalyst-app:
- `codemod/new-route/index.js`
- `codemod/new-route/template.js`
- `codemod/new-route/transformers/createReducer.js`
- `codemod/new-route/transformers/createRoute.js`
- `codemod/new-route/utils.js`
- `scripts/cli.cjs`

## Acceptance criteria

- [x] `ansis` added as a dependency to both `catalyst-core` and `create-catalyst-app` package.json; `picocolors` removed
- [x] All 14 files migrated to `ansis`'s API (verified equivalent per import pattern: default/namespace/named, ESM + CJS — not a blind find/replace)
- [x] `terminalProgress.js` re-verified: per-line `console.log` calls, not single-call multi-line coloring — low risk, though not independently testable locally (Babel-transpiled at build; CI is first real check)
- [x] `cli.cjs` (CJS) — confirmed `ansis`'s CJS `require()` works identically to `picocolors`'s
- [x] No visible change to existing colored output — confirmed CCA's old `picocolors@0.1.0` colors were genuinely functional pre-migration (not a silent no-op), so no regression

## Parent
Story: #343

## PR
#390

mayankmahavar1mg:
## Summary

Closes #389. Migrates `picocolors` → `ansis` across catalyst-core and create-catalyst-app (15 files, 1-line-per-file swap).

Found while designing #360 (output modes): `picocolors` doesn't re-apply ANSI color codes across `\n` line breaks within a single colored multi-line string — colors can leak/break across lines. `ansis` handles this correctly (verified empirically, see below) and is comparably fast in realistic benchmarks. This matters because #360's `--verbose`/`--debug` modes need a boxed, multi-line, colored error format.

Doing this as its own PR first, ahead of #360, since it's unrelated in scope to the output-modes feature and touches files #360 doesn't otherwise need to touch.

## Verification

- `ansis`'s default-import (`import pc from`), namespace-import (`import * as pc from`), and named-export (`import { red, bold } from`) patterns all tested against the real installed package — confirmed equivalent to `picocolors`'s API, including under CJS `require()`.
- Multi-line re-coloring behavior confirmed directly: `ansis.red("line1\nline2\nline3")` correctly wraps each line in its own color codes; this is the entire justification for the swap.
- `ansis@4.3.1`'s `engines` requirement is `>=14` — no Node version floor change for consumers.
- CCA pins an old `picocolors@^0.1.0` (vs. catalyst-core's `^1.0.0`) — checked this wasn't silently broken before assuming parity: installed `0.1.0` directly, confirmed `red()`/`green()`/`cyan()`/`bold()` all produce real ANSI codes under `FORCE_COLOR=1`. No prior-broken-output regression risk.

## Not verified locally

`native/terminalProgress.js` is Babel-transpiled at the package's `prepare` script (`babel src/native --out-dir dist/native`), and this repo has no installed `node_modules` for that package locally (`npx babel` here resolves to the unrelated legacy `babel` CLI package). Confirmed via `node --check` that the pre-existing `picocolors` import fails identically (not a regression from this change), but the actual transpiled output isn't verified locally — **CI's `prepare` step is the first real check on this file.**

## Test plan

- [x] `node --check` on all non-transpiled files — passes
- [x] `ansis` import patterns (default/namespace/named, ESM + CJS) verified against installed package
- [x] Multi-line color re-application verified
- [x] `engines` requirement checked (no floor change)
- [x] CCA's `picocolors@0.1.0` colors confirmed genuinely functional pre-migration
- [ ] `terminalProgress.js` Babel transpile — relying on CI


mayankmahavar1mg:
Rebased onto the updated `feature/386-runtime-web` (which itself moved from `story/6-error-handling` onto `feature/372-cca` — see #391). Base branch (`feature/386-runtime-web`) is unchanged; only the underlying commits shifted. Content of this PR is unchanged.

Issue #392, opened 2026-08-06
catalyst build/serve fails with ERR_REQUIRE_ESM in native/bridge/errors.js during offline manifest generation


mayankmahavar1mg:
## Description

`catalyst build` fails for any example app with native bridge hooks (confirmed with `examples/useai`) during offline manifest generation:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module .../dist/errors/index.js from .../dist/native/bridge/errors.js not supported.
```

`src/native/bridge/errors.js` uses ESM `import { ERROR_CODES, createError } from "../../errors/index.js"`. Vite's SSR module runner (invoked from `generateOfflineManifest.js`'s `loadAppRoutes`) loads this file via a CJS-style `require()` path during route discovery, which Node rejects since the target is an ESM module.

Confirmed via `npm run test:example -- examples/useai --only web`, reproduced identically on a clean `feature/372-cca` checkout with no other changes — this is pre-existing, not introduced by #360/#386/#389.

## Impact

- `catalyst build` fails on the offline manifest generation step for any app importing native bridge hooks (`useGoogleSignIn`, etc. — anything importing `native/bridge/hooks.js`)
- `catalyst serve` fails as a result (depends on a successful build)
- `catalyst start` (dev mode) is unaffected — passes

## Suggested fix directions (not investigated in depth)

- Have `generateOfflineManifest.js`'s route-loading path use dynamic `import()` instead of whatever triggers the CJS require, or
- Split `native/bridge/errors.js`'s dependency on `errors/index.js` so the native bridge subtree doesn't cross the ESM/CJS boundary Vite's SSR runner can't cross, similar to how `src/native`'s CJS-only subtree already avoids this for `buildAppAndroid.js`/`buildAppIos.js`.

## Parent
Story: #343 (found during #360 example-app verification, but affects build/serve broadly — not necessarily error-handling-story scoped; flagging under Story 6 since it was found there, happy to move if it belongs elsewhere)

PR #387, opened 2026-08-06
feat(errors): add RUNTIME-WEB-xxx error category and wire handler.jsx SSR catch sites

mayankmahavar1mg:
## Summary

Closes #386. Adds the `RUNTIME-WEB-xxx` error category (promised in the RFC #155 discussion but never built) and wires all 5 raw `console.error` sites in `handler.jsx`'s SSR request handler into the error registry.

This closes a gap found while checking `examples/` apps: any error thrown during SSR — the user's own code, a third-party npm library, or catalyst-core's own pipeline — was previously logged as a raw, unformatted error with no code, no docs link, and no consistent structure.

## Design notes

- `RUNTIME-WEB-001..004` map to render / fetcher / serverSideFunction / request-handling stages. All 4 are origin-unknown wrappers (we can't tell from the caught error whether it's app code, a dependency, or catalyst-core) — see the amendment comment on #386 for the full reasoning, including why this deviates from the `-000`/`-001+` numbering convention used by `BUNDLE`/`IOS`/`ANDROID`/`AI`.
- New `wrapSSRError(stage, err)` helper added alongside (not replacing) `wrapForeignError`, since the build-stage model (one code, hardcoded "Build failed" message) didn't fit SSR's 4 distinct stages with no named upstream tool.
- Unchanged: HTTP status/response behavior at each catch site, and the error object passed to `safeCall(onRenderError/onFetcherError/onAppServerSideError/onRequestError, ...)` — hooks still receive the original unwrapped error.
- Stack traces preserved via a second `console.error(error)` call per site rather than embedding the stack in `formatError`'s output (tested the combined-arg form first — Node appends the inspected Error onto the same line as the "Docs:" line, which reads badly).

## Not included

`errors/index.json` is **not** regenerated in this PR. It's a full-overwrite artifact built from whatever `ERROR_DEFINITIONS` are visible on the branch that regenerates it — this branch doesn't have the AI/CCA/bridge codes from #370/#371/#372 (they're on `feature/372-cca`, stacked separately). Regenerating here would produce a partial index that could clobber their entries at merge time. **Needs a regenerate pass (`node packages/catalyst-core/src/errors/generateDocs.js`) after this and the #370/#371/#372 stack are both merged into `story/6-error-handling`.**

## Note

[...]

Issue #386, opened 2026-08-06
Add RUNTIME-WEB-xxx error category and wire handler.jsx SSR catch sites


mayankmahavar1mg:
## Description

`packages/catalyst-core/src/server/renderer/handler.jsx` (SSR request handler) is not wired into the error registry shipped in #359. Found during a follow-up audit prompted by checking the `examples/` apps: any error thrown inside app-level code during SSR — the user's own code, a third-party npm library, or catalyst-core's own fetcher/render pipeline — is indistinguishable today, because all 5 catch sites in `handler.jsx` log via raw `console.error` instead of going through `createError`/`CatalystError`:

- `onError` inside `renderToPipeableStream` (line ~302)
- render document catch (line ~312)
- `serverFetcher` execution catch (line ~406)
- `App.serverSideFunction` catch (line ~425)
- top-level request-handling catch (line ~444)

This is also the `RUNTIME-WEB-xxx` category the RFC #155 discussion committed to (see the maintainer's reply to @TrinaChoudhury, item 5: "Split categories: RUNTIME-WEB-xxx: SSR errors, API failures, server crashes, client JavaScript errors") but the category was never added to `registry.js` — only `RUNTIME-NATIVE-xxx` exists today.

## Design notes

- Add a new `RUNTIME-WEB-xxx` category to `registry.js`, and a runtime stage to `wrapForeignError` (`STAGE_WRAPPER_CODE`/`STAGE_UPSTREAM_NAME`) alongside the existing `BUNDLE`/`IOS`/`ANDROID` build stages.
- Unlike build-stage wrapping, the runtime catch sites cannot distinguish "user's own app code" vs. "a third-party library the user installed" vs. "catalyst-core's own pipeline code" — the caught `error` object doesn't carry that provenance. Following the same precedent as `BUNDLE-000`/`IOS-000`/`ANDROID-000` (origin-agnostic wrapper, original error preserved as `cause`, never reinterpreted), the wrapper code here means "something failed during this SSR stage, origin unknown" rather than attempting attribution.
- Wrapping is for the `console.error` *display* path only — must not change:
  - which HTTP status gets rendered (each catch site has different fallback behavior: 404, `error.status_code`, or just calling the request-error hook)
  - the `error` object passed to `safeCall(onFetcherError/onAppServerSideError/onRenderError/onRequestError, ...)` — user hooks should keep receiving the original error, not the wrapped one

[...]

PR #393, opened 2026-08-06
feat(errors): implement three output modes (default/verbose/debug)

### LINKED ISSUE #360: Implement three output modes (default/verbose/debug)

mayankmahavar1mg:
## Summary

Closes #360. Implements the three RFC #155 §3.2 output modes (default/`--verbose`/`--debug`) across every module system in the codebase: catalyst-core (ESM), create-catalyst-app (CJS), and `src/native`'s CJS-only buildApp scripts.

## What changed

**ESM core** (`errors/index.js`, `scripts/scriptUtils.js`): `formatError(err, mode, env)` — mode is a parameter, not module state, so `WebBridge.js` (browser, no `process.argv`) is unaffected and keeps getting default-mode output. `resolveOutputMode(argv, env)` uses the existing `--inspect`-style bare-flag convention (`serve.js`), not `arrayToObject` (which drops valueless flags). Wired into `build.js`, `server/utils/validator.js`, `handler.jsx` (5 sites from #386/#391), with `serve.js`/`start.js` forwarding the resolved mode to the spawned `expressServer.js` child via `CATALYST_OUTPUT_MODE`.

**CCA** (`create-catalyst-app/scripts/errors.cjs`, `cli.cjs`): mirrors the ESM implementation in CJS. 5 genuine error call sites wired; the CCA-009 `.gitignore`-exists notice (informational, not an error) intentionally left at default-mode-only.

**Native** (`buildAppAndroid.js`/`buildAppIos.js`): neither called `formatError` before — hand-written single-line strings, since `src/native` is CJS-only and can't reach the ESM `errors/index.js`. Added a shared `src/native/buildErrorFormat.js` rather than duplicating the box/mode logic. iOS's error path bypasses `progress.log` (which prefixes every call with an icon/color, incompatible with a multi-line box) in favor of plain `console.error`, matching Android.

**Also fixed**: `registry.js#getDefinition()`'s fallback previously mislabeled unrecognized codes as `RUNTIME-NATIVE` — now surfaces as `Category: UNKNOWN`, motivated by the new `Category:` line added to verbose/debug output (not in the original RFC example, but genuinely useful and already-available data).

## Also in this stack (separate PRs, prerequisite work found along the way)

- #391 (was #387, closed/reopened after a branch restack — see below): `RUNTIME-WEB-xxx` category + `handler.jsx` SSR wiring
- #390: `picocolors` → `ansis` migration, needed because `picocolors` doesn't correctly re-apply color across `\n` line breaks in multi-line strings (this PR's boxed verbose/debug output)

## Branch restack note

This branch (and #386/#389 beneath it) was rebased from `story/6-error-handling` onto `feature/372-cca`, since CCA's `errors.cjs` (needed for this PR's CCA scope) didn't exist on the old base. GitHub refused to retarget PR #387's base in place once it detected a stack relationship, so #387 was closed and replaced by #391 with the corrected base. No PR content changed, only the base branch.

## Filed separately (found during verification, out of scope here)

- #392: pre-existing `ERR_REQUIRE_ESM` bug in `native/bridge/errors.js` during `catalyst build`'s offline manifest generation — reproduced identically on a clean `feature/372-cca` checkout with none of this stack's changes, confirmed not caused by this work.

## Test plan

- [x] All 3 modes manually verified against real `CatalystError`/`CCAError` instances (ESM + CJS), including multi-line cause chains and unknown-code category fallback
- [x] `resolveOutputMode` precedence tested: argv flag beats inherited env var, invalid env values ignored, all combinations
- [x] `ansis` multi-line color re-application confirmed working inside the box format
- [x] `catalyst buildApp` (no platform suffix) confirmed to forward `--verbose`/`--debug` to both native scripts via `bin/catalyst.js`'s `scriptArgs`
- [x] `npm pack --dry-run` on create-catalyst-app confirmed clean with the new `ansis` dependency
- [x] Ran `npm run test:example -- examples/useai --only web` against this branch — `start` (dev server) passes; `build`/`serve` fail on a pre-existing, unrelated bug (#392), confirmed via baseline comparison on `feature/372-cca` with none of this stack's changes
- [ ] eslint not run locally (no resolvable `eslint.config.js` from these directories) — CI is the first real lint gate

### DIFF
--- a/packages/catalyst-core/src/server/renderer/handler.jsx
+++ b/packages/catalyst-core/src/server/renderer/handler.jsx
@@ -33,6 +33,21 @@ import createStore from "@catalyst/template/src/js/store/index.js"
 import { SsrRequestProvider } from "../../web-router/components/SsrRequestContext.jsx"
 import { getManifest, getAssetManifest } from "../manifestCache.js"
 import { wrapSSRError, formatError } from "../../errors/index.js"
+import { resolveOutputMode, getDebugEnvInfo } from "../../scripts/scriptUtils.js"
+
+// Resolved once at module load — spawned via serve.js/start.js, which
+// forward the mode as CATALYST_OUTPUT_MODE (no argv available at this
+// level; see scriptUtils.js#resolveOutputMode).
+const outputMode = resolveOutputMode(process.argv, process.env)
+// Debug mode's boxed output already embeds the full stack trace (see
+// errors/index.js#formatDebug), so only default/verbose need the original
+// error printed separately to avoid losing the stack.
+const logSSRError = (stage, error) => {
+    const wrapped = wrapSSRError(stage, error)
+    const debugEnv = outputMode === "debug" ? getDebugEnvInfo() : undefined
+    console.error(formatError(wrapped, outputMode, debugEnv))
+    if (outputMode !== "debug") console.error(error)
+}
 
 const DEFAULT_SAFE_AREA_INSETS = { top: 0, right: 0, bottom: 0, left: 0 }
 
@@ -300,8 +315,7 @@ const _renderMarkUp = async (
                 },
 
                 onError(error) {
-                    console.error(formatError(wrapSSRError("RENDER", error)))
-                    console.error(error)
+                    logSSRError("RENDER", error)
                     safeCall(onRenderError, { req, res, store, error })
                     cleanupSafeArea()
                     tail.destroy(error)
@@ -311,8 +325,7 @@ const _renderMarkUp = async (
         })
     } catch (error) {
         cleanupSafeArea()
-        console.error(formatError(wrapSSRError("RENDER", error)))
-        console.error(error)
+        logSSRError("RENDER", error)
         safeCall(onRenderError, { req, res, store, error })
         return Promise.reject(error)
     }
@@ -406,8 +419,7 @@ async function _handler(req, res) {
                     )
                 }
             } catch (error) {
-                console.error(formatError(wrapSSRError("FETCHER", error)))
-                console.error(error)
+                logSSRError("FETCHER", error)
                 safeCall(onFetcherError, { req, res, store, error })
 
                 if (res.headersSent) return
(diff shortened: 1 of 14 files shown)


mayankmahavar1mg:
#review

packages/catalyst-core/src/server/renderer/handler.jsx
```diff
@@ -33,6 +33,21 @@ import createStore from "@catalyst/template/src/js/store/index.js"
 import { SsrRequestProvider } from "../../web-router/components/SsrRequestContext.jsx"
 import { getManifest, getAssetManifest } from "../manifestCache.js"
 import { wrapSSRError, formatError } from "../../errors/index.js"
+import { resolveOutputMode, getDebugEnvInfo } from "../../scripts/scriptUtils.js"
+
+// Resolved once at module load — spawned via serve.js/start.js, which
+// forward the mode as CATALYST_OUTPUT_MODE (no argv available at this
+// level; see scriptUtils.js#resolveOutputMode).
+const outputMode = resolveOutputMode(process.argv, process.env)
```
trinachoudhury1mg: why not pick it directly from process.env.CATALYST_OUTPUT_MODE ?

mayankmahavar1mg: you're right, the comment right above even says argv isn't available at this level, so the argv checks inside resolveOutputMode are dead code here. will simplify to read CATALYST_OUTPUT_MODE straight from env.

mayankmahavar1mg:
fixed in 22cc1310 — handler.jsx now passes an empty argv to resolveOutputMode() since this process never sees the parent CLI's flags. found the same dead-argv pattern in server/utils/validator.js and fixed that too for consistency.

PR #399, opened 2026-08-10
fix(core): upgrade react-router-dom v6 to v7.18.2 (CVE fixes)

mayankmahavar1mg:
## Summary
- Bumps `react-router-dom`/adds `react-router` from `^6.16.0` to `^7.18.2`, resolving three moderate CVEs left explicitly out of scope in #375 (open redirect via backslash, XSS via open redirects, SSR hydration constructor injection).
- Fixes the SSR `StaticRouter` import (`react-router-dom/server` was removed in v7 — moved to `react-router`).
- Fixes a v7-specific SSR module-runner break: `react-router-dom`'s Node resolution condition defaults to CJS, which broke Vite's `ssrLoadModule()` (used by `generateOfflineManifest.js`) trying to derive named exports from it. Switched catalyst-core's internal imports to `react-router` directly, whose top-level index resolves to ESM under Vite SSR and already re-exports everything used here (`createBrowserRouter`, `RouterProvider`, `matchRoutes`, hooks, `StaticRouter`).
- Extends the same `react-router-dom` → `react-router` swap to every remaining call site: `MetaTag.jsx`, the `.d.ts` type barrel and type-only references, and 11 files in `examples/test-video-hook-poc` that imported `react-router-dom` directly (bypassing catalyst-core's barrel).
- Adds `react-router` as a direct dependency in `test-video-hook-poc`'s `package.json` since its source now imports it directly rather than only transitively through catalyst-core.

## Out of scope / explicitly not done
- **Not a v8 upgrade.** A prior session attempted v6→v8 directly and aborted it: v8 requires a React version bump (≥19.2.7, this repo is pinned at 19.0.0) and splits `react-router-dom` into `react-router` + `react-router/dom` imports, which would be a breaking change to catalyst-core's own public router API surface. This PR stays on v7.18.2, which needs neither.
- Two pre-existing, unrelated bugs surfaced during manual browser testing of the example app and are **not fixed here** (confirmed via `git blame` — neither file touched by this branch beyond the import swap):
  - `examples/test-video-hook-poc/src/js/containers/Home/components/Theme.js`: SSR/CSR hydration mismatch from reading `localStorage`/`matchMedia` synchronously in a `useState` initializer.
  - `FilePickerPanel.js` / `useFilePicker.js`: an unrelated "Maximum update depth exceeded" loop.

## Test plan
- [x] `semgrep scan` (security-audit, secrets, owasp-top-ten) on all changed files — 0 findings
- [x] `trivy fs` on root `package-lock.json` — 0 vulnerabilities (confirms CVE-2026-53666/53668/53669 resolved)
- [x] Clean-cache `npm run test:example -- examples/test-video-hook-poc --only web`: build + offline-manifest generation pass
- [x] Live SSR smoke test: HTTP 200, full rendered page, `vendor-router` chunk loaded, no duplicate-React/invalid-hook-call errors
- [x] Manual browser test (dev mode, incognito + normal) clicking through all tabs — no router-related errors; the two hydration/loop issues found are pre-existing and unrelated (see above)
- [x] Cherry-picked cleanly onto `fix/security-dashboard-issues`'s base with no conflicts and a clean build, confirming compatibility with that PR's changes


packages/catalyst-core/package.json
```diff
@@ -49,7 +49,8 @@
         "react-helmet-async": "^1.3.0",
         "react-redux": "^9.2.0",
         "react-refresh": "^0.14.0",
-        "react-router-dom": "^6.16.0",
+        "react-router": "^7.18.2",
```
trinachoudhury1mg, line 53: Ideally we should move to peer dependency format. 
ie
react/react-dom/redux/react-router versions supported should be defined. 
Also, if we are defining react-router for the time being here, any reason to add it to the test app as well?

PR #403, opened 2026-08-13
Security vuln fix

D-Agam:
- Updated the tata1mg/router version
- Updated the react-router-dom version
- These are necessary to resolve the react-router-dom vuln issues reported in the psp-pfizer service


mayankmahavar1mg (approved):
Outdate pr
merging to unblock security cases

Issue #340, opened 2026-08-03
Story: Test coverage and CI enforcement


mayankmahavar1mg:
## Goal

Back the API-stability promise of 1.0 with automated test coverage and CI enforcement. Currently `packages/catalyst-core` has no unit tests (only a fixture smoke test), and CI is advisory only — no branch protection blocks merge on failing checks.

## Tasks

- [ ] Write unit tests for the router
- [ ] Write unit tests for serverFetcher/clientFetcher
- [ ] Write unit tests for the SSR handler
- [ ] Add branch protection requiring CI to pass before merge

## Parent

Epic: #329

mayankmahavar1mg:
# Story #340 — reframed scope draft

Reframe per Mayank's direction: the story's real payload is contract-testing the
error system just built in Story 6, not generic router/fetcher/handler unit tests.
Enforcement must be pipeline-level, not hook-level (contributors can bypass hooks).

## Layer 1 — Error-system contract tests (new core of the story)

1. **Registry shape invariants** (`packages/catalyst-core/src/errors/registry.js`)
   - Every `ERROR_CODES` value has a matching `ERROR_DEFINITIONS` entry, and vice versa
   - Codes are unique
   - Code string matches `CATEGORY-NNN` format
   - Code's category prefix matches `def.category`
   - All required fields present and non-empty (category, defaultMessage, defaultDetails,
     suggestedAction — `recoverable` pending the drop landing via restack)
   - → This is the actual "new errors must follow the guideline" enforcement: a
     malformed new code fails CI with no human review needed.

2. **Docs-drift check**
   - Run `generateDocs()` into a temp dir, diff against committed `errors/**/*.md` +
     `errors/index.json`
   - Fails if someone edits registry.js without regenerating docs, or hand-edits a
     generated .md directly
   - Directly catches the "contributor changes core, doesn't realize docs are generated"
     mistake

3. **API behavioral tests** (`packages/catalyst-core/src/errors/index.js`)
   - `createError`: override semantics (message/details/recoverable/suggestedAction/cause)
   - `wrapForeignError`: per-stage code mapping, unknown-stage throw, named-code vs
     numeric-exit-code branch (the `hasNamedCode` check)
   - `wrapSSRError`: stage → code mapping, unknown-stage throw
   - `wrapProviderError`: cause construction, provider/status attached
   - `getDefinition`: UNKNOWN-category fallback for unrecognized codes
[...]

PR #409, opened 2026-08-18
fix: react-helmet-async peer conflict, offline-manifest .jsx resolution, classic JSX runtime trap (#408)

### LINKED ISSUE #408: catalyst serve breaks on 0.3.0-beta.5: react-helmet-async unresolvable + offline-manifest .jsx extension bug
## Summary

Upgrading a consumer app to `catalyst-core@0.3.0-beta.5` breaks `catalyst serve` (production) with an opaque, swallowed error. Root-caused to two independent framework bugs plus a logging gap that made diagnosis much harder than it should have been. Reproduced from a clean `npm install` on an unmodified internal consumer app by bumping only the `catalyst-core` version — no other app changes needed to trigger it.

## Symptom

```
🚀 Starting production server...
[catalyst-core/ai] catalyst-ai not installed — AI routes unavailable
unhandledRejection in Catalyst {"code":"ERR_MODULE_NOT_FOUND"}
```

No module name, no stack trace, no line number. Server exits. (`catalyst-ai` is a red herring — that log line is expected/harmless when the optional `catalyst-ai` package isn't installed; `mountAIRouter` in `expressServer.js` catches its absence and returns cleanly.)

## Bug 1 — `react-helmet-async` is left external in the SSR build but isn't hoistable to the consumer's `node_modules`

`catalyst-core`'s `sharedViteConfig.ssr` does not add `react-helmet-async` to `noExternal`, so Vite's SSR build leaves it as a bare external `import` in the generated `build/server/index.js`. At runtime, Node resolves that bare specifier starting from the **consuming app's** `node_modules`, not catalyst-core's.

`react-helmet-async@1.3.0`'s `peerDependencies` cap at React 18 (`^16.6.0 || ^17.0.0 || ^18.0.0`), while `catalyst-core@0.3.0-beta.5` ships/depends on React 19. Because of that peer mismatch, npm's installer will not hoist `react-helmet-async` to the app root — it stays nested under `catalyst-core/node_modules/react-helmet-async`, invisible to the app's own module resolution. Result: `ERR_MODULE_NOT_FOUND` at server startup in production.

**Verified fix:** adding `react-helmet-async` to `ssr.noExternal` in `packages/catalyst-core/src/vite/vite.config.js` bundles it into the SSR output instead of leaving it external — confirmed working end-to-end (clean build + `catalyst serve` returns HTTP 200) with **zero consumer-app changes**.

Longer-term note: `react-helmet-async@1.3.0` is unmaintained/archived and has no React 19 peer range. Worth evaluating the maintained fork, or migrating off it toward React 19's native document-metadata support (`<title>`/`<meta>` in the render tree).

## Bug 2 — `generateOfflineManifest.js` hardcodes `.js` extension instead of resolving the app's actual route-utils file

`packages/catalyst-core/src/router/ClientRouter.js`, `ServerRouter.js`, and `server/renderer/handler.jsx` all import the app's routes via the extensionless, alias-based specifier:

```js
import { ... } from "@catalyst/template/src/js/routes/utils"
```

which lets Vite's normal resolver try `.js`, `.jsx`, etc.

But `packages/catalyst-core/src/scripts/generateOfflineManifest.js` instead does:

```js
const routesModule = await vite.ssrLoadModule(path.join(appRoot, "src/js/routes/utils.js"))
```

— a literal filesystem path with a hardcoded `.js` suffix, bypassing extension resolution entirely. Any app whose file is `routes/utils.jsx` (the common case for a file containing JSX, and the actual convention in this app) fails here with:

```
Error when evaluating SSR module .../src/js/routes/utils.js: Failed to load url .../src/js/routes/utils.js. Does the file exist?
```

**Verified fix:** switching this call to the same extensionless bare specifier the other three files already use (`@catalyst/template/src/js/routes/utils`) resolves correctly against `.jsx`, with the original filename left untouched.

## Bug 3 (related, lower severity) — offline-manifest's throwaway Vite server doesn't load the app's config/env

While reproducing, any module reachable from `routes/utils` that reads `process.env.*` at import time (e.g. `JSON.parse(process.env.SOME_CONFIG_KEY)`) throws inside `generateOfflineManifest.js`'s standalone `createViteServer()` instance, because that instance doesn't load `config/config.json` into `process.env` the way the real dev/prod server does. This is app-adjacent but worth a framework-level look since any app whose routes tree transitively imports a module with import-time env reads will hit it.

## Bug 4 — swallowed error obscures all of the above

`packages/catalyst-core/src/server/expressServer.js`:

```js
function safeStringify(err) {
    try {
        return JSON.stringify(err)
    } catch (e) {
        console.log("error in safeStringify", e)
        return err
    }
}
process.on("unhandledRejection", (err) => console.log("unhandledRejection in Catalyst", safeStringify(err)))
```

`JSON.stringify` on a native `Error` object only serializes enumerable own properties (typically just `code` on a `SystemError`), dropping `message` and `stack` entirely. This is why the real failure (Bug 1) took significant time to diagnose — the log gave no indication of which module or file was missing.

**Suggested fix:** log `err?.stack || err?.message || err` instead of `safeStringify(err)`.

## Repro steps

1. Take any consumer app pinned to an older `catalyst-core` ref where `react-helmet-async` hoists cleanly and `routes/utils.jsx` (not `.js`) is used.
2. Bump `catalyst-core` to `^0.3.0-beta.5` and run `npm install`.
3. `npm run build && npm run serve`.
4. Observe `ERR_MODULE_NOT_FOUND` with no further detail (Bug 4 masking Bug 1), or — once Bug 1 is worked around — the offline-manifest `.jsx` resolution failure (Bug 2).

## Suggested labels

`bug` (regression in `catalyst serve` production path introduced between the pre-0.3.0-beta.5 baseline and 0.3.0-beta.5)

mayankmahavar1mg:
## Summary

Fixes tata1mg/catalyst-core#408 — `catalyst serve` breaking on `0.3.0-beta.5` for any consumer app on React 19.

## Changes

1. **`react-helmet-async` peer conflict** — added a root-level `overrides` entry forcing its `react`/`react-dom` peer range to `19.0.0`. Without this, npm nests `react-helmet-async` invisibly under `catalyst-core/node_modules` instead of hoisting it, and the SSR build's external (unbundled) import of it then fails to resolve at runtime with `ERR_MODULE_NOT_FOUND`.
2. **`generateOfflineManifest.js`** — switched from a hardcoded literal `src/js/routes/utils.js` path to the same extensionless alias specifier (`@catalyst/template/src/js/routes/utils`) that `ClientRouter.js`/`ServerRouter.js`/`handler.jsx` already use. The old hardcoded path failed for any app using `utils.jsx` (the common case for a file containing JSX).
3. **`jsxInJsPlugin`** — now passes `jsx: "automatic"` to esbuild, matching `@vitejs/plugin-react`'s runtime for `.jsx`/`.tsx` files. Previously it defaulted to esbuild's classic runtime, silently requiring `React` to stay in scope in any `.js` file with JSX — inconsistent with every `.jsx` file in the same app, and an easy trap once #2 above pushes someone to rename `.jsx` → `.js`.
4. **Swallowed `unhandledRejection` logging** — `expressServer.js` now logs `err.stack`/`err.message` instead of `JSON.stringify(err)`, which drops both fields for real `Error` objects. This masked the actual failure and is why #408 took so long to diagnose.
5. **`examples/useai`** — removed the redundant `@tata1mg/router` dependency (catalyst-core already re-exports all of `react-router` plus `RouterDataProvider`/`MetaTag`). It was pulling in a duplicate `react-router`/`react-router-dom` copy that caused a separate SSR crash (`UNSAFE_invariant` in `useMatch`) once the app was on a clean install.

## Verification

All four fixes verified end-to-end against `examples/useai` with a **fully clean install** (`rm -rf node_modules package-lock.json`, then `sync-core.js`):
- Root `npm install`: no ERESOLVE, `react-helmet-async` hoists correctly
- Build with `utils.jsx` (not renamed): offline manifest generates cleanly (confirms fix #2)
- Build with `utils.js` and no `React` import (the exact trap from #408): builds and serves cleanly (confirms fix #3)
- `catalyst build && catalyst serve`: **HTTP 200** on `/` and a nested route, no `ERR_MODULE_NOT_FOUND`, no `ReferenceError: React is not defined`

Not in scope here (tracked separately in tata1mg/catalyst-core#340): `sync-core.js`'s `--package-lock=false` install masks dependency-resolution regressions like this one by never re-resolving an existing `node_modules`, which is why `test:example` didn't catch #408 originally.


mayankmahavar1mg:
## CodeRabbit review summary

Ran `coderabbit review --agent --base main` against this branch. 1 finding, fixed and verified.

| # | File | Severity | Finding | Fix | Rating |
|---|------|----------|---------|-----|--------|
| 1 | `packages/catalyst-core/src/server/expressServer.js` | Minor | `describeError` read `err.stack`/`err.message` **outside** the `try` block — a Proxy or an object with a throwing getter for either property would make the handler itself throw inside `process.on('unhandledRejection', ...)`, defeating the point of the diagnosability fix in this PR | Moved all property access inside `try`; catch now returns a fixed `"<unserializable rejection>"` string instead of re-touching the same risky value via `String(err)` | ⭐⭐⭐⭐⭐ Valid, correctly scoped, fixed |

**Verification after fix:** unit-checked `describeError` against a normal `Error`, a plain string, a plain object, and a deliberately throwing `Proxy` — all four return correctly, no throw. Rebuilt and served `examples/useai` end-to-end afterward — still `HTTP 200`, no regressions.

Fixed in `f4087d95`, pushed to this branch.

Issue #421, opened 2026-08-19
setMetaData doesn't apply on client-side navigation — title clears instead of updating


mayankmahavar1mg:
## Symptom

Client-side (SPA) navigation via `<Link>`/`useNavigate` does not apply the target route's `setMetaData()` — the `<title>` tag goes **empty**, not stale. Reproduced live: navigating from `/` (title "Home") to `/about` via a client-side link click leaves `document.title === ""`, not "Home" (stale) and not "About" (correct).

Confirmed reproducible via `examples/catalyst-core-test` (the Playwright fixture app, see #412) — its `tests/main.spec.js` has had this exact case commented out since before this story, with the note "Currently breaking - needs to be fixed in catalyst":

```js
// Currently breaking - needs to be fixed in catalyst
// test("setMetaData on CSR", async ({ page }) => {
//   await page.goto("http://localhost:3005/");
//   await page.getByRole("link", { name: "About" }).click();
//   await expect(page).toHaveTitle("About");
// });
```

Re-enabling it and running against current `main` reproduces the failure:

```
Error: expect(page).toHaveTitle(expected) failed
Expected: "About"
Received: ""
```

SSR (`setMetaData on SSR` test, initial page load) works correctly — this is CSR-navigation-specific.

## Suspected root cause

`packages/catalyst-core/src/web-router/components/MetaTag.jsx`:

```jsx
useEffect(() => {
    const mergedMetaTags = getMetaData(matchedRoutes, routeData)
    if (Array.isArray(mergedMetaTags) && mergedMetaTags.length > 0) {
        setMetaTags(mergedMetaTags)
    }
    return () => deleteHeadTagsByDataAttribute("catalyst")
    // eslint-disable-next-line react-hooks/exhaustive-deps
}, [location])
```

and `packages/catalyst-core/src/web-router/utils/metaDataUtils.jsx#getMetaData`:

```js
matchedRoutes?.forEach?.((match) => {
    const setMetaData = match?.route?.component?.setMetaData
    ...
})
```

[...]

Issue #415, opened 2026-08-19
CI enforcement: new workflow jobs + required status checks


mayankmahavar1mg:
## Goal

Make CI the actual non-bypassable gate — today `.github/workflows/node.js.yml` only runs `pull_request -> main` (never on the stacked task/story/epic branches where review actually happens), has no lint job, and the "Restrict Main" ruleset enforces zero required status checks (verified via `gh api repos/tata1mg/catalyst-core/rulesets` — it only blocks deletion/force-push/creation/update). Local hooks (husky, the `secure-commit` skill) are bypassable via `--no-verify`, a clone that skips `npm install`, or any fork PR — they can't be the actual enforcement mechanism.

## Scope

- [ ] Add `lint` job — oxlint + oxfmt --check (independent of PR #396/#378 landing; coordinate configs if #396 merges first)
- [ ] Add `test-js-unit` job — see #411
- [ ] Add `test-js-integration` job — see #412
- [ ] Add `test-android-unit` job — see #413
- [ ] Add `test-ios-unit` job — see #414
- [ ] Widen `node.js.yml`'s trigger beyond `pull_request: branches: [main]` so it runs on task/story/epic branches too
- [ ] Add required-status-check rules to the "Restrict Main" ruleset for all of the above — required immediately, no advisory-first grace period (any breakage a new check surfaces gets fixed in the same PR that adds it)
- [ ] Extend `.husky/pre-commit` with a fast subset (lint-staged + new contract tests + gitleaks) as local convenience/fast-feedback only — not a replacement for CI enforcement
- [ ] Add an ESLint/oxlint rule banning raw `throw new Error(...)` and bare `console.error(...)` inside `packages/catalyst-core/src` (exempting the errors module itself), mechanically enforcing the `createError`/wrap* convention for new code

## Parent

Story: #340

mayankmahavar1mg:
Note from #423 (TypeScript conversion for #411's test suites, verified via fresh-install check): none of the current `node.js.yml` jobs actually invoke `npm run test:unit` (the `tsc --noEmit` + `vitest run` path) — `test-catalyst-core.sh` only runs the Playwright fixture's own `npm run test`. So the new framework-level test suites (catalyst-core, catalyst-ai, create-catalyst-app, mcp_v2 — 94 tests + type-checks) aren't wired into CI on any branch yet.

[...]

Issue #411, opened 2026-08-19
Error-system contract tests (Tier 1 — JS unit)


mayankmahavar1mg:
## Goal

Framework-level (Tier 1) unit tests proving catalyst-core's error system (Story 6) is internally consistent — registry shape, docs generation, and behavioral contracts of `createError`/`wrap*`/`formatError`. These test the framework's own machinery in isolation, not an app built with it (see #340's Tier 2 sub-issues for that).

## Scope

Set up vitest for `packages/catalyst-core` (no JS test runner exists in the repo today) and `packages/create-catalyst-app`, tests under each package's own `test/` directory.

- [ ] **Registry shape invariants** (`packages/catalyst-core/src/errors/registry.js`): every `ERROR_CODES` value has a matching `ERROR_DEFINITIONS` entry and vice versa; codes unique; `CATEGORY-NNN` format; category prefix matches `def.category`; required fields non-empty (category, defaultMessage, defaultDetails, suggestedAction)
- [ ] **Docs-drift check**: run `generateDocs()` into a temp dir, diff against committed `errors/**/*.md` + `errors/index.json` — fails if registry.js changes without regenerating docs
- [ ] **API behavioral tests** (`packages/catalyst-core/src/errors/index.js`): `createError` override semantics, `wrapForeignError` per-stage mapping + unknown-stage throw + named-code vs numeric-exit-code branch, `wrapSSRError` stage mapping, `wrapProviderError` cause construction, `getDefinition` UNKNOWN-category fallback, `formatError` in all 3 modes
- [ ] **CCA mirror parity**: structural test asserting `create-catalyst-app/scripts/errors.cjs` stays in sync with the shared registry's shape/category conventions

## Parent

Story: #340

mayankmahavar1mg:
## Progress: all 4 checklist items done

Branch: `feature/340-error-contract-tests` (cut from `story/6-error-handling`). 62 tests, all passing via `npm test --workspaces --if-present`.

- **Registry shape invariants** — 13 tests (`packages/catalyst-core/test/errors/registry.test.js`)
- **Docs-drift check** — 3 tests (`docsDrift.test.js`)
- **API behavioral tests** — 23 tests (`index.test.js`) covering `createError`/`wrapForeignError`/`wrapSSRError`/`wrapProviderError`/`getDefinition`/`formatError`
- **CCA mirror parity** — 23 tests (`packages/create-catalyst-app/test/errors.test.cjs`)

[...]

PR #418, opened 2026-08-19
test(errors): Layer 1 contract tests for the error system (#411)

mayankmahavar1mg:
## What

Was originally scoped to Layer 1 (framework-level JS unit / contract tests) for issue #340's test-coverage story. Since opening, this branch became the collapse point for the full test-coverage stack — #422, #426, #429, and #434 have all been merged into it (their own PRs closed, see comments on each) — so it now resolves the JS, Android, and iOS Tier 1 work plus CI enforcement together, not just the original JS-only scope.

Fixes #411
Fixes #412
Fixes #413
Fixes #414
Fixes #415

First JS test runner in the monorepo (vitest, chosen for ESM-native + fast + fits the mixed ESM/CJS package layout here). **91 tests, all passing**, run via:

```
npm run test:unit
```

## Coverage

| Package | Tests | What |
|---|---|---|
| `catalyst-core` | 39 | error registry shape invariants, docs-drift (generated `.md`/`index.json` vs committed), `createError`/`wrap*`/`formatError` behavioral contract |
| `create-catalyst-app` | 23 | CCA error mirror (`errors.cjs`) structural + behavioral parity with the shared registry conventions |
| `catalyst-ai` | 23 | `route.js` internals — AI_CONFIG parsing, the `MODEL_NAME_RE` injection guard, all 4 provider usage normalizers, `throwProviderError` |
| `catalyst-mcp` (`mcp_v2`) | 6 | `explain_error` — packaged-path + monorepo-fallback resolution, own-property lookup safety |

Every suite was mutation-tested at least once during development (a real bug deliberately reintroduced, confirmed the test catches it, reverted) rather than just trusted on green output.

## Notable implementation decisions

- **`generateDocs.js`**: `errorsDir` is now a parameter (default: the real repo-root `errors/` dir) instead of a hardcoded module constant, so the docs-drift test can regenerate into a scratch temp dir without mutating the committed tree. Fully backward-compatible — verified zero output drift.
- **`catalyst-ai/route.js`**: attached pure-logic internals as `router._internal` (an extra property on the exported Express router — safe, ignored by Express, doesn't change the `require("catalyst-ai/route")` contract) since the router itself was the only export. `_internal` is test-only, not a supported public API.
[...]

PR #422, opened 2026-08-19
refactor: consolidate apps/ into examples/, pin Node 22 (#412)

### LINKED ISSUE #412: Wire existing Playwright fixture into CI (Tier 2 — JS integration)
## Goal

Application-level (Tier 2) proof that catalyst-core works as actually consumed — not just that its internals are correct in isolation.

## What already exists

`apps/catalyst-core-test` + `scripts/test-catalyst-core.sh` already do this correctly: `npm pack` the local `catalyst-core`, install it for real (not symlinked) into the fixture app, build, and run 6 Playwright tests (SSR render, server/client fetchers, `setMetaData`, server middleware, env vars). This is the honest "does it work when someone actually builds with it" check — it's just never been wired into any CI workflow.

## Scope

- [ ] Add a `test-js-integration` job to `.github/workflows/node.js.yml` running `npm run core:test` (`scripts/test-catalyst-core.sh`)
- [ ] Confirm it runs headless/CI-clean (Playwright browsers installed via `npx playwright install --with-deps` or similar)
- [ ] Re-enable or explicitly document the currently-commented-out `setMetaData on CSR` test in `apps/catalyst-core-test/tests/main.spec.js` (marked "Currently breaking - needs to be fixed in catalyst") — file separately if it's a real bug, don't silently leave it commented once CI is gating on this suite

## Parent

Story: #340

mayankmahavar1mg:
## What

Follow-up to #411 for issue #340 / #412 — Tier 2 (application-level integration test) wiring, plus a structural cleanup requested along the way.

## Correction to #412's original scope

I was wrong in #412's issue body: `scripts/test-catalyst-core.sh` was **already wired into CI** (`.github/workflows/node.js.yml`'s `catalyst-core` job, running on the Playwright container image) — not an unwired gap as originally claimed. Posted a correction on #412. The real, narrower gap (trigger scope — `node.js.yml` only fires on `pull_request → main`, never on stacked task/story/epic branches) belongs to #415, not duplicated here.

## What this PR actually does

1. **Consolidated `apps/` into `examples/`** (Mayank's call — one top-level dir for example/fixture apps, not two). `git mv apps/catalyst-core-test examples/catalyst-core-test`, 30 clean renames, updated the 3 places referencing the old path (`scripts/test-catalyst-core.sh`, `README.md`, a comment in `packages/catalyst-core/vitest.config.js`). Verified `examples/`'s own tooling (`test-all.sh`, `sync-core.js`) requires an explicit app-path argument, so nothing auto-discovers and accidentally sweeps in the moved fixture.

2. **Pinned Node 22** (`.nvmrc` + root `"engines": {"node": ">=22"}`, per "use node 22 always"). While verifying the `apps/` → `examples/` move by actually running the fixture test locally, hit a real failure: local Node 20.4.0 can't `require()` a `.js` file containing `import` syntax, which broke Playwright's config loader. Confirmed this is environment-specific, not a repo bug — CI's Playwright container image ships Node 24.x and has never hit it. Fixed properly (pinned the floor) rather than working around it. Also bumped the `docs` CI job off a hardcoded `node-version: "20"` to `node-version-file: ".nvmrc"` so they can't drift apart again.

   **Two things to confirm rather than assume:**
   - `engines` is scoped to the **private root** `package.json` only, not the published `catalyst-core`/`create-catalyst-app` packages — pinning real consumers of those packages to Node 22 would be a breaking-change decision this PR doesn't make unilaterally.
   - `engines` without npm's `engine-strict` setting only **warns**, doesn't hard-block. Left as a warning for now — say if you want `engine-strict=true` in a root `.npmrc` to make it enforced.

3. **Re-verified the fixture end-to-end** under Node 22: build succeeds, `playwright.config.js` loads, all 6 Playwright tests pass. `npm run test:unit` (#411's suites) still green — 94 tests total.

4. **Closed #412's last open checklist item**: re-enabled the commented-out `setMetaData on CSR` test to check whether it's still broken. It is — confirmed live (title clears to `""` on client-side navigation, not stale). Reverted the test back to commented-out (not this PR's job to fix a framework bug), filed **#421** with a root-cause hypothesis (traced into `MetaTag.jsx`/`metaDataUtils.jsx`), and left a pointer comment in the test file instead of an undocumented `// Currently breaking`.

## Test plan

```bash
git checkout feature/412-playwright-ci
nvm use   # picks up .nvmrc (22)
npm install
npm run test:unit          # 94 tests
npm run core:test           # 6 Playwright tests against examples/catalyst-core-test
```

## Parent

Story: #340 (sub-issue #412)


mayankmahavar1mg:
This merged automatically once its content became fully contained in #418 (feature/340-error-contract-tests) — noting here for the record: #418 stays open as the collapsed integration branch for the full stack, with fixes made along the way (vitest min-release-age gate pin, etc.). All CI is green there.

Issue #427, opened 2026-08-20
react-helmet-async peer-deps conflict with React 19 masked by stale lockfile


mayankmahavar1mg:
## Context

While fixing #415's \`test-js-unit\` CI job, regenerating \`package-lock.json\` from scratch (needed to capture the missing \`@rollup/rollup-linux-x64-gnu\` platform entry — see PR #426) surfaced a real peer-dependency conflict that a plain \`npm install\` refuses to resolve:

\`\`\`
Could not resolve dependency:
peer react@"^16.6.0 || ^17.0.0 || ^18.0.0" from react-helmet-async@1.3.0
node_modules/react-helmet-async
  react-helmet-async@"^1.3.0" from catalyst-core@0.3.0-beta.3

Found: react@19.0.0
\`\`\`

\`react-helmet-async@1.3.0\`'s peer range only covers React 16–18; this repo is on React 19. The **previously committed** lockfile tolerated this silently because \`npm ci\` installs exactly what's locked without re-resolving peers — the conflict only surfaces on a fresh \`npm install\`. PR #426 worked around it with \`--legacy-peer-deps\` purely to regenerate the lockfile's platform-specific optional-dependency entries; that flag is not a real fix.

## Task

- [ ] Check for a react-helmet-async release/fork with React 19 peer support (e.g. \`react-helmet-async@2.x\` if it exists, or a maintained fork)
- [ ] If none exists, evaluate replacing it (react 19 has built-in support for rendering \`<title>\`/\`<meta>\` tags directly in components, which may remove the need for react-helmet-async entirely)
- [ ] Once resolved, regenerate \`package-lock.json\` with a plain \`npm install\` (no \`--legacy-peer-deps\`) to confirm the conflict is actually gone, not just silenced
- [ ] Re-verify CI (lint, test-js-unit, test-js-integration, create-catalyst-app, catalyst-ai, docs) all pass with the clean lockfile

## Parent

Story: #340, discovered while working #415

PR #426, opened 2026-08-20
ci: add lint job, wire test:unit into CI, widen PR trigger (#415)

mayankmahavar1mg:
## What

Implements the "implementable now" subset of #415's CI-enforcement checklist. Stacks on top of `feature/412-playwright-ci`.

- Widens `node.js.yml`'s `pull_request` trigger beyond `main` to `epic/**`, `story/**`, `task/**`, `feature/**` — CI now actually runs where stacked-PR review happens.
- New `lint` job running `oxlint` (see rule details below).
- New `test-js-unit` job running the root `npm run test:unit` (tsc --noEmit + vitest, 94 tests across catalyst-core, catalyst-ai, create-catalyst-app, mcp_v2). Closes the gap flagged in a comment on #415 after PR #423 — no job previously exercised this path on any branch.
- Renamed the existing `catalyst-core` job to `test-js-integration` for clarity (already runs the Playwright fixture suite from #412).
- New custom oxlint JS plugin (`tools/oxlint-plugins/catalyst-errors.mjs`) banning `throw new Error(...)`/other bare built-in Error subclasses in `packages/catalyst-core/src` (excluding `src/errors`) — oxlint has no `no-restricted-syntax` equivalent, so this needed a real plugin.
- Enabled oxlint's built-in `no-console` rule in the same scope (log/warn/info/debug/trace still allowed).
- Extended `.husky/pre-commit` with a fast local subset (lint-staged + `test:unit` + `gitleaks protect --staged`) — convenience only, **not** a replacement for CI enforcement (bypassable via `--no-verify`).
- Fixed a genuine TS2309 diagnostic in `catalyst-ai/src/route.d.ts` (`export =` can't coexist with other top-level `export`s under this module setting) found by oxlint.

## Why warn, not error, on the two new rules

They currently flag **292 pre-existing violations across 44 legacy files** in `src/`. Fixing those is real, separate work — tracked in #425, where the rules get flipped to `"error"` once clean. This PR only adds the mechanism.

## Deliberately out of scope

- `test-android-unit` / `test-ios-unit` jobs — blocked on #413/#414 (not started).
- `oxfmt --check` — no oxfmt config exists yet; its defaults disagree with the repo's current Prettier formatting on ~213/291 files. Tracked in #424, to land after #396.
- Required-status-check rules on the "Restrict Main" ruleset — a live, non-PR-gated repo-settings change. Deferred until these jobs prove green and #413/#414 exist.

## Known risk to watch in CI

[...]

Issue #425, opened 2026-08-20
Fix 292 pre-existing throw-new-Error/console.error violations in catalyst-core/src, flip oxlint rules to error


mayankmahavar1mg:
## Context

#415 added two oxlint rules scoped to \`packages/catalyst-core/src\` (excluding \`src/errors\`):

- \`catalyst-errors/no-throw-new-error\` (custom plugin, \`tools/oxlint-plugins/catalyst-errors.mjs\`) — bans \`throw new Error(...)\` and other bare built-in Error subclasses in favor of \`createError()\`/\`wrapError()\`/\`wrapForeignError()\`/\`wrapSSRError()\`/\`wrapProviderError()\`
- \`no-console\` (allowing log/warn/info/debug/trace, banning \`console.error\`)

Both are set to \`"warn"\` rather than \`"error"\` because they currently flag **292 violations across 44 files**, all pre-existing and unrelated to #415's scope. Run \`npm run lint:oxlint\` (drop \`--quiet\`, or just \`npx oxlint .\`) to see the current list.

## Task

- [ ] Go through the 44 flagged files under \`packages/catalyst-core/src/native/\`, \`src/scripts/\`, \`src/server/\`, \`src/otel.js\`, \`src/sentry.js\` and convert each raw \`throw new Error(...)\` to the appropriate \`createError\`/\`wrap*\` call, and each bare \`console.error(...)\` to whatever the equivalent structured-logging path ends up being (there's currently no dedicated "print an error" helper exported from \`src/errors\` — decide whether one is needed, or whether \`formatError()\` + the existing wrap functions cover it)
- [ ] Flip both rules from \`"warn"\` to \`"error"\` in \`.oxlintrc.json\` once clean
- [ ] Remove the now-unnecessary \`--quiet\` flag from the \`lint:oxlint\` script / CI step if no other warn-level noise remains (oxlint's default ruleset — unused-vars, unicorn/*, typescript/* — may still produce warnings; re-evaluate at that point whether those should also be addressed or explicitly ignored)

## Parent

Story: #340, follow-up to #415

Issue #417, opened 2026-08-19
Fix sync-core.js stale node_modules masking dependency-resolution regressions


mayankmahavar1mg:
## Goal

`examples/sync-core.js` — the mechanism `test:example`'s default (non-`--catalyst-version`) path uses to get local `catalyst-core` source into an example app — installs via `npm install --no-save --package-lock=false "${tarballPath}"` against the example's **existing** `node_modules`, without ever deleting it first. Because `--package-lock=false` skips lockfile reconciliation, this only resolves/places the package being synced — it does not re-resolve the rest of the dependency tree, so stale transitive deps from a previous install silently persist even if they'd no longer be resolvable under the current `package.json`.

## How this let #408 slip through

`react-helmet-async` (peer-capped at React <=18) was hoisted correctly in `examples/useai/node_modules` from before catalyst-core's React 19 bump introduced the conflict — a stale leftover carried forward by every subsequent `sync-core.js` run since. `test:example` was therefore exercising a `node_modules` state a fresh install could never produce. Confirmed: `node examples/sync-core.js --only web` still hard-fails with `ERESOLVE ... peer react@"^16.6.0 || ^17.0.0 || ^18.0.0" from react-helmet-async@1.3.0` when starting from a clean `node_modules`/lockfile.

## Scope

- [ ] Add a clean-install mode to `sync-core.js` (flag, or always-on for `test:example` runs) that removes the example's `node_modules` + `package-lock.json` before installing, so each run exercises real dependency resolution
- [ ] Verify this would have caught #408 at the point `react-helmet-async` was added without a React-19-compatible peer range

## Origin

Filed while root-causing #408; full writeup posted as a comment on #340.

## Parent

Story: #340

PR #429, opened 2026-08-20
test: Android Mockito + CI, wire iOS XCTest into CI (#413, #414)

mayankmahavar1mg:
## What

Implements #413 (Android Tier 1 unit tests + Mockito) and #414 (iOS Tier 1 XCTest wired into CI), plus a substantial follow-up coverage-raising effort on the Android side that grew well beyond the original scope of #413. Stacks on top of `feature/415-ci-enforcement` (#426).

**Android package-wide JVM unit test coverage: 16.66% → 61.30%** over this PR (see breakdown below). 518 tests, 0 failures.

---

### #414 — iOS: wire XCTest into CI

New `test-ios-unit` CI job: `xcodebuild test` on `macos-latest` against the existing 13-file XCTest suite (no new mocking library, per the issue). Needed several iterative real-CI-driven fixes — none reproducible locally:

- **Simulator destination**: replaced a hardcoded device name (broke as soon as GitHub rotated its preinstalled simulator set) with a runtime picker via `xcrun simctl list devices available -j` (structured JSON, talks only to CoreSimulator) instead of text-scraping `xcodebuild -showdestinations` (flaky/slow on a cold SPM cache, failed silently under `2>/dev/null`). Excludes "e"-suffixed budget-line models.
- **Explicit pre-boot step**: `simctl boot` + `bootstatus -b` before `xcodebuild test`, so simulator boot time (~82s) is isolated and visible in the job log.
- **`-parallel-testing-enabled` OFF**, not on — caused CPU-contention flakiness on the shared runner without a real speed win.
- **Root-caused two real flaky tests**, not CI-side band-aids:
  - `BootTimingUtilityTests.testEdgeCase_ConcurrentLogging` was failing at exactly its timeout (10s, then 60s after a bump) — real bug: dispatched work on `DispatchQueue.global(qos: .background)`, which the OS can starve indefinitely under CPU contention. Fixed by switching to `.userInitiated`.
  - Confirmed no other test in the suite uses `.background` QoS.
- `xcbeautify --quieter`, SPM packages cached via `actions/cache` (measured no real build-time difference — the ~8.4 min step is dominated by compiling project code, not fetching deps).
- 3 consecutive green runs confirmed in real CI.

### #413 — Android: Mockito + Tier 1 unit tests

- `mockito-core` 5.23.0, `mockito-kotlin` 6.3.0, `kotlinx-coroutines-test` 1.10.2 as `testImplementation`. Narrowed the existing `resolutionStrategy.eachDependency` override so it no longer force-downgrades `kotlin-reflect` (needed at ≥2.1.20 by mockito-kotlin 6.x) while keeping the stdlib pinned.
- New `test-android-unit` CI job: `./gradlew testDebugUnitTest` on `ubuntu-latest` + JDK 17.
- First Mockito-based test in this template: `OfflineCacheServiceTest.kt`.
- Separately: fixed issue #433 — `create-catalyst-app` CI failing on npm's min-release-age gate (pinned vitest to 4.1.9) plus a second, independent npm arborist `edgesOut` crash in `create-release-sandbox.js`'s own lockfile-less install (fixed via a targeted `overrides` entry), verified with a positive/negative control and a full local smoke-test pass.

### Jacoco coverage reporting + gate scoping

- Added a `jacocoTestReport` Gradle task (`build(android): add Jacoco JVM unit test coverage reporting`) — real, measured line coverage instead of just pass/fail counts.
- **Coverage is intended to become a CI merge gate** (block PRs that drop the number below the last known value). The raw package-wide number initially measured ~16.66% against *all* 6,844 authored lines — including ~2,900 lines that are structurally untestable without Robolectric or a real device (real `WebView`/`Activity`/`View` construction: `CustomWebView`, `MainActivity`, `NativeBridge`'s instance body, security alert UI, camera session binding).
- Investigated Robolectric as a way to cover that Tier-2 code: it works (tests run, ~1.9s one-time bootstrap cost per JVM fork, then ~0.02s/test), **but JaCoCo cannot attribute coverage through Robolectric's sandboxed classloader** — a confirmed, unresolved 10-year-old upstream limitation ([robolectric/robolectric#2230](https://github.com/robolectric/robolectric/issues/2230), [#5575](https://github.com/robolectric/robolectric/issues/5575)). A coverage number that can't move for real work is useless as a gate, so this was reverted rather than shipped.
- Instead, scoped the jacoco `excludes` to the same Tier 1 (testable) / Tier 2 (framework-bound) split already established on iOS with the `CatalystCoreLogic` package split (#432) — classified every main-source file via a dedicated pass (not filename-guessing), excluded each Tier 2 file's compiled class + lambdas from the gate's class tree. This is a **measurement correction, not new coverage**: same lines covered, denominator drops from 6,844 to 3,951 honestly-testable lines. Reported number moved from 16.66% to 28.85% with zero new tests.
- Where a Tier 2 file has an already-tested companion object (`NativeBridge.parseAndValidateMessage`, `PluginBridge.parseRequest`), only the instance class is excluded, not the companion — worked around Gradle's Ant-`exclude()` not supporting `!` negation by naming the excluded class directly (`NativeBridge.class`, not a wildcard that would also catch `NativeBridge$Companion.class`).

### Coverage-raising batches (28.85% → 61.30%)

Systematic pass through every 0%/low-coverage Tier 1 file, several run as parallel subagents in isolated git worktrees then merged sequentially:

| Area | Result |
|---|---|
| MetricsMonitor (real production code, runs in every build type — not just debug) | 0% → 75.4% |
| WebCacheManager | replaced a vanity test file that never instantiated the class → 35.4% |
| NotificationUtils | extended `buildNotification`/style branching → 49.5% |
| DownloadUtils | 0% → 23.7% |
| FridaDetector + SecurityCheckManager | 0% → 64.6% / 75.4% (real bound-socket technique to force detection-positive deterministically) |
| PluginBridge instance methods | 21.9% → 69.9% |
| 7 small camera/design/util files (BarcodeDetector, HoldController, TorchController, ViewfinderMapper, DesignTokens, NetworkMonitor, SafeAreaUtils) | 0% → 37–100% each |
| ZoomController, DeviceInfoUtils, SecurityCheckScheduler | 0% → 74–89% each |
| PushNotificationUtils (noFcm stub) + `Utils.kt`'s `matchesCachePattern` | 0% → 100% / 89% |
| FileSizeRouterUtils — extended `processFile`'s 3 transport branches | 31.7% → 84.9% |
| OfflineCacheService — extended to cover manifest-driven routing (pre-writing `manifest.json` to unlock real route-match branches) | 31.2% → 50.2% |
| `FilePickerOptions`/`GoogleSignInOptions` (NativeBridge) — widened `private` → `internal` (visibility only, no logic change) to make two standalone, framework-free JSON-parsing classes testable | NativeBridge companion 17.2% → 79.3% |

Also removed one stale vanity test (`DeviceInfoUtilsTest.kt` root-package version, which asserted against hand-built JSON literals and never called the class), superseded by a real test in the correct package.

**Deliberately left out of scope**, documented per-file: real `java.net` network paths (`fetchAndCacheResourceBlocking`, `revalidateInBackground`, manifest refresh) — this project's standing stance is not to mock `java.net` deeply enough that the test verifies the mock instead of the code; `Build.*` static-final fields (not interceptable by Mockito — confirmed empirically, affects `EmulatorDetector`, parts of `DeviceInfoUtils`); `CustomWebView` (checked for the same "extract pure logic" opportunity NativeBridge had — no equivalent found, every URL-parsing-looking method is an instance method needing a real `WebView`).

**Real production bugs found and fixed (with sign-off before touching production code, per this repo's review norms)**: `/status` endpoint returning 406 to any real client (no `ContentNegotiation` installed) and `serverScope` being a permanently-cancelled `val` `CoroutineScope` after the first `stopServer()`, silently killing periodic file cleanup on every server restart thereafter — both in `FrameworkServerUtils.kt`.

## Verification

- `./gradlew testDebugUnitTest`: 518/518 passing, 0 failures, 0 errors (confirmed fresh, not cached).
- `./gradlew jacocoTestReport`: 2422/3951 lines = 61.30%, verified via the generated XML, not eyeballed from the HTML report.
- iOS `test-ios-unit`: 3 consecutive green CI runs.
- Android `test-android-unit`: green in CI throughout.

## Not in this PR

- Robolectric/Maestro for the remaining ~2,800 Tier 2 (framework-bound) lines — explicitly evaluated and deferred (see gate-scoping section above).
- The CI coverage *gate* itself (failing a PR when coverage drops) — this PR only makes the number honest; wiring the actual gate is separate follow-up work.
- Issue #425 (292 legacy oxlint violations), #427 (react-helmet-async peer conflict), #416 (native Tier 2 emulator/simulator tests), #417 (sync-core.js stale node_modules) — tracked separately.


mayankmahavar1mg:
## CodeRabbit CLI review

Ran `coderabbit review --agent --base origin/feature/415-ci-enforcement` against this branch (43 files changed). The service dropped the WebSocket connection twice on earlier attempts (large diff); a third run completed cleanly and reviewed all 43 files. Findings below, verified against current code before listing (severities re-rated where CodeRabbit's own label didn't hold up).

| # | Severity | File | Finding | Verified? |
|---|---|---|---|---|
| 1 | ~~Critical~~ → Info | `BridgeUtilsTest.kt` | `whenever(webView.evaluateJavascript(...)).thenThrow(...)` used on a void Android method instead of `doThrow(...).whenever(...)` | **Not a real bug** — ran the test standalone, it passes. mockito-kotlin's `whenever` handles this correctly here. CodeRabbit's general Mockito rule doesn't hold for this setup. |
| 2 | Minor (real) | `FrameworkServerUtils.kt` | `serverScope` (reassigned in `startServer()`, read in `stopServer()`/coroutine launches) isn't `@Volatile` — no guaranteed cross-thread visibility of the reassignment | **Valid.** This is the `val`→`var` fix from earlier this session; the volatility gap is a legitimate follow-up. Production code — did not apply myself, flagging for your call. |
| 3 | Minor (real) | `NotificationUtilsTest.kt:244` | `assertNotNull(callbackInvoked)` on a non-null `Boolean` is tautological — never actually verifies the callback fired | **Valid bug in the test.** Should be `assertTrue(callbackInvoked)`. |
| 4 | Minor (real) | `FrameworkServerUtilsTest.kt` | `setUp()` doesn't defensively `stopServer()` before tests, only `tearDown()` does — a prior test class leaving the JVM-wide singleton server running could poison this class's `isRunning()==false` assumptions | **Valid**, same class of shared-state hazard fixed elsewhere this session (`FileSizeRouterUtilsTest`, `WebCacheManagerTest) — just missed here. |
| 5 | Minor (real) | `FileSizeRouterUtilsTest.kt:137` | Test named "...for a file over the max server size" actually asserts a **100-byte** file stays on `BRIDGE_BASE64` — name doesn't match what it tests | **Valid** naming/clarity issue, not a logic bug. |
| 6 | Minor (real) | `FileSizeRouterUtilsTest.kt:319` | `FrameworkServerUtils.startServer(...)` return value isn't checked before proceeding — if the real bind fails on a busy CI runner, the test fails confusingly instead of skipping | **Valid** — should use `org.junit.Assume.assumeTrue(...)` around the bind, matching `FridaDetectorTest`'s pattern. |
| 7 | Info (real, low-priority) | `PerfEventBufferTest.kt:41` | `configure(true)` test comment assumes the debug variant without asserting `BuildConfig.DEBUG` | **Valid but low-impact** — `testDebugUnitTest` effectively guarantees the debug variant already. |
| 8 | Minor (real) | `FridaDetectorTest.kt` | `bindOrSkip(port)` doesn't actually skip on bind failure — it rethrows as `IllegalStateException`, so the name is misleading and the negative ("clean env") tests aren't protected by an `Assume`-style tolerance either | **Valid** — helper name doesn't match behavior. |
| 9 | ~~Major~~ → Unconfirmed | `catalyst-ai/package.json` | Claims vitest should be `4.1.11` instead of `4.1.9` for a security fix | **Could not confirm.** The CVE I found matching this shape (CVE-2026-53633, CVSS 9.8, Browser Mode RCE) is fixed as of **4.1.8** per multiple advisories — 4.1.9 (pinned deliberately this session for an unrelated npm arborist bug, see `89f6ea11`) is already patched. No advisory found requiring 4.1.11 specifically. Flagging as unconfirmed rather than acting on it. |
| — | Minor (iOS, out of my scope) | `NativeBridgeTests.swift:446-453` | Test-owned `DispatchQueue.main.asyncAfter` timers used instead of waiting on the actual JS permission callback | Surfaced on all 3 review attempts — recurring finding, likely real, but iOS is outside what I touched this session. Flagging for the iOS side owner. |

### Net: 2 test-file bugs worth fixing now (#3, #6), 3 lower-priority test robustness gaps (#4, #5, #8), 1 real production follow-up for your call (#2), 1 unconfirmed dependency claim (#9), 1 false positive (#1), 1 iOS item outside this session's scope.

Not fixing anything yet — flagging here first per request. Let me know which to act on.

mayankmahavar1mg:
Fixed in `aff3e556`:

| # | Item | Status |
|---|---|---|
| 2 | `serverScope` needs `@Volatile` | ✅ Fixed |
| 3 | Tautological `assertNotNull` | ✅ Fixed |
| 4 | Missing defensive `stopServer()` in setUp | ✅ Fixed |
| 5 | Misleading test name | ✅ Fixed |
| 6 | Missing `Assume` around server bind | ✅ Fixed |
| 7 | Debug-variant assumption | ✅ Fixed (now an explicit `Assume`) |
| 8 | `bindOrSkip` doesn't actually skip | ✅ Fixed (renamed + real skip added for the negative test) |
| 1 | BridgeUtilsTest "critical" | Not applied — false positive, test passes as-is |
| 9 | vitest 4.1.11 claim | Not applied — unconfirmed, 4.1.9 already patches the CVE I could find matching this shape |
| — | iOS `NativeBridgeTests.swift` timer-based waits | Not applied — outside this session's scope, flagging for iOS owner |

518/518 tests passing, 0 skipped, 0 failures. Coverage unchanged at 61.30% (these were quality fixes, not new coverage).

mayankmahavar1mg:
Closing — this branch's content has been merged into #418 (feature/340-error-contract-tests), along with fixes made along the way (vitest min-release-age gate pin, CodeRabbit review fixes, an iOS test flake resolved). #418 stays open as the collapsed integration branch; all CI is green there.

Issue #416, opened 2026-08-19
Native Tier 2: scaffolded-app emulator/simulator integration tests (Android + iOS)


mayankmahavar1mg:
## Goal

Application-level (Tier 2) proof for native platforms — analogous to #412's JS Playwright fixture, but for Android/iOS. Currently no equivalent exists: `androidProject`/`iosnativeWebView` are the *template source* catalyst-core copies into scaffolded apps, and their existing test suites (see #413/#414) only cover Tier 1 (unit-level, in-place, no real app/device/emulator).

## Why this is separate from #413/#414

Emulator/simulator-based CI is materially more expensive (runner cost, flakiness, setup complexity) than headless JVM/XCTest unit runs. Landing Tier 1 for all three platforms first, unblocking CI enforcement (#415) sooner, then coming back for this.

## Scope (rough — needs its own design pass before implementation)

- [ ] Android: scaffold a real app from `androidProject` via `create-catalyst-app`, run `androidTest` (instrumented) on a GitHub Actions emulator runner, verify bridge commands actually round-trip through a real WebView
- [ ] iOS: same, scaffold from `iosnativeWebView`, run on an `xcodebuild test -destination "platform=iOS Simulator,..."` simulator
- [ ] Decide what scenarios are worth the cost — likely a small, high-value set (e.g. one bridge round-trip per RUNTIME-NATIVE error category actually triggered on-device), not exhaustive coverage duplicating Tier 1

## Parent

Story: #340

mayankmahavar1mg:
Attempted in PR #453 (now closed). **Emulator/simulator instrumented coverage does not fit per-PR CI:**

- `android-instrumented-coverage` timed out at **17m25s** — `connectedDebugAndroidTest` hangs/crawls because `MainActivity`/`CustomWebView` instrumented tests each init a full WebView subsystem on a headless serverless emulator. Locally: one `CustomWebView` construction test = 524s. AVD snapshot caching doesn't help (cost is test execution, not boot).
- `ios-app-uitests` (parallel macOS job) passed but at **8m07s** vs the ~6-min budget; both iOS jobs ran ~2x slow that run.
- The working assumption that instrumented tests slot into 'free parallel headroom' was wrong — this is structurally a 10-17+ min workstream and would grow CI wall-clock (~9 min) materially.

**This issue needs a different approach** (or deprioritization). Options not pursued: nightly-only scheduled workflow (out of scope per requester).

[...]

PR #434, opened 2026-08-21
feat(ios): split CatalystCore into UIKit-free CatalystCoreLogic target (#432)

### LINKED ISSUE #432: iOS: split CatalystCore SPM target so pure-logic tests can run via swift test (no simulator)
## Context

Investigating why `test-ios-unit` CI is slow (part of #340/#414 work), profiling showed the job's ~11-13 min wall time is dominated by the `xcodebuild test` build step (~8.4 min build vs ~1s of actual per-test execution once running), not simulator boot (already reduced to ~82s via #429) or test count itself.

## Finding

7 of 13 files in `iosnativeWebViewTests/` have zero UIKit/WebKit/SwiftUI imports and only depend on `@testable import CatalystCore`. Tests like these, if the code under test has no UIKit/WebKit dependency, can run via `swift test` directly on the host (no simulator boot, no app bundle install, no XCTest IPC harness) — a well-documented pattern shows this being ~60x faster for equivalent suites (25s -> 0.4s) since it's just running a plain executable.

`CatalystCore` (`packages/catalyst-core/src/native/iosnativeWebView/Package.swift`) is a single monolithic SPM target. Verified locally: attempting `swift test` against it fails immediately because other files in the same target (`CameraPreviewView.swift`, `WebView.swift`, etc.) `import UIKit` directly, and UIKit does not exist on macOS — `swift test` builds the whole target as one compilation unit, so it can't skip UIKit-dependent files just because a given test doesn't touch them.

## Verified safe split boundary

A dependency-mapping pass over all 45 files in `Sources/Core/`, plus direct verification of the build generators, established the exact safe boundary:

**Safe to move to a new `CatalystCoreLogic` target (13 files, zero circular-dependency risk):**
- `Sources/Core/Camera/VideoStreamState.swift`
- `Sources/Core/Camera/VideoStreamStateMachine.swift`
- `Sources/Core/Notifications/NotificationModels.swift`
- `Sources/Core/Protocols/NotificationHandlerProtocol.swift`
- `Sources/Core/Utils/NotificationHandlerProvider.swift`
- `Sources/Core/Utils/NullNotificationHandler.swift`
- `Sources/Core/Utils/BootTimingUtility.swift`
- `Sources/Core/Utils/CacheManager.swift`
- `Sources/Core/Utils/FrameworkServerUtils.swift`
- `Sources/Core/Utils/NetworkMonitor.swift`
- `Sources/Core/Utils/URLWhitelistManager.swift`
- `Sources/Core/Constants/CatalystConstants.swift` (required by 3 of the above; UIKit import is `#if canImport(UIKit) ... #elseif canImport(CoreGraphics)`-guarded purely for `CGFloat` in the checked-in file)
- `Sources/Core/Constants/ConfigConstants.swift` (same guard in the checked-in file — but see critical gap below)

**Excluded — real circular-dependency risk, must stay in the UI target:**
- `Sources/Core/Camera/HoldController.swift` — references `BarcodeDetector` (imports AVFoundation)
- `Sources/Core/Plugins/CatalystPlugin.swift` — protocol method takes a `PluginBridgeContext` param (imports WebKit/UIKit), and `PluginBridge.swift` depends back on `CatalystPlugin` — mutual entanglement
- `Sources/Core/Plugins/GeneratedPluginIndex.swift` — auto-generated at build time by `pluginComposerIos.js`, references arbitrary UIKit-unconstrained plugin classes; entire `Plugins/` directory stays in the UI target

**Test files — 6 of 7 move, 1 stays:**
- Move: `BootTimingUtilityTests.swift`, `CacheManagerTests.swift`, `ConfigMappingTests.swift`, `FrameworkServerUtilsTests.swift`, `NotificationHandlerTests.swift`, `URLWhitelistManagerTests.swift`
- Stays with the simulator-based suite: `ScreenSecureManagerTests.swift` — directly references `UIScene.willDeactivateNotification`/`didActivateNotification` (no macOS equivalent)

**No new SPM dependencies needed for the logic target**: `GoogleSignIn` is only imported by `BridgeCommandHandler.swift`, `JSONSchema` only by `BridgeMessageValidator.swift` — both stay UI-only.

## Considered and rejected: N-way split (Camera/WebView/Plugins as separate targets, for parallel CI)

There are zero Camera tests in the suite, and all 6 UI-domain-bound test files depend on the same monolithic UI code — a "Camera CI job" would have nothing to test, and each UI-domain shard would still re-pay the full ~8 min build + 82s simulator boot to run a fraction of the ~3 min of actual test execution. The real, available parallelism is the 2-target split: a fast `swift test` job (seconds, no simulator) running concurrently with the existing simulator-based job as two separate CI jobs.

## Critical constraints found during planning (all verified directly, not assumed)

1. **`Package.swift` is regenerated, not static.** `packages/catalyst-core/src/native/buildIos/plugins.js` dynamically rewrites `Package.swift` on every build where a config hash changes, and already has precedent for a second conditional target (`CatalystNotifications`). The checked-in `Package.swift` is only the fallback stub for when Xcode opens the project directly. **The split must be implemented in this generator**, not just hand-edited into the static file, or it's silently overwritten on the next build.

2. **`ConfigConstants.swift`'s generator produces a BROKEN (unguarded) file.** The checked-in file has the safe `canImport(UIKit)` guard, but `buildIos/config.js` (via `generateConfigConstants`) generates it with only `import Foundation` and bare `CGFloat` — no guard at all. Any real build overwrites the safe file with this broken one, breaking `swift build`/`swift test` on macOS. **Must fix `buildIos/config.js` to add the same guard** as part of this work.

3. **Missing `platforms` declaration will break `swift test`.** Confirmed locally: `swift test` failed with `error: the library 'CatalystCore' requires macos 10.13, but depends on the product 'GoogleSignIn' which requires macos 10.15`. `CacheManager.swift` also needs `CryptoKit` (macOS 10.15+). Both `Package.swift` variants need `platforms: [.iOS(.v17), .macOS(.v13)]` (or similar) added.

4. **Stale/broken `CatalystCoreTests` testTarget already exists** in the checked-in `Package.swift` (`path: "Tests/CatalystCoreTests"`, directory never created) — confirmed this causes `swift build` to fail outright (`invalid custom path`). Must be removed and replaced with a correctly-pathed `CatalystCoreLogicTests` target.

## Recommended approach: spike first

This investigation already hit four different "should work" assumptions that failed on first real attempt (across CI timeout/simulator-picker/QoS work and this split investigation itself). Recommend: implement with just `BootTimingUtility.swift` + its test first, wire both Package.swift variants with all the fixes above, and confirm `swift build --product CatalystCoreLogic` + `swift test --filter BootTimingUtilityTests` both work standalone (no simulator, no accidental full-UIKit-target build) *and* survive a real `buildAppIos`/`buildIos` regeneration cycle — before moving the remaining 12 files.

## Scope note

This is a real package-structure refactor (source file moves across 2 targets, 2 build-generator fixes, Xcode project surgery to deduplicate test compilation, new CI job) — not a CI config tweak. Full step-by-step implementation plan available on request. Tracked here as a follow-up to #413/#414, alongside #416/#417/#433.

mayankmahavar1mg:
## Summary

Splits `CatalystCore`'s SPM target into a UIKit-free `CatalystCoreLogic` nested package + the existing `CatalystCore` (UI/WebKit) target, so pure-logic tests can run via `swift test` with zero simulator dependency — closes #432.

## What changed

- **Architecture**: `CatalystCoreLogic` is a genuinely separate nested SPM package at `iosnativeWebView/CoreLogic/`, not a second target inside the existing package — a target-inside-package approach was tried first and found not to work (`swift test`/`swift build` at a package root always resolve/build every target regardless of `--target`/`--filter` scoping). See commit `7535c03b` for the full investigation.
- **11 of the originally-planned 12 files moved** into `CoreLogic/Sources/CatalystCoreLogic/`. `CacheManager.swift` stays in `CatalystCore` — discovered at compile time that it depends on `OfflineCacheService`/`CatalystPerf`, both genuinely `WKWebView`-bound (same disqualification class as `HoldController`/`CatalystPlugin` from the original dependency scan).
- **Generator updated**: `buildIos/plugins.js` emits the `.package(path: "CoreLogic")` dependency; `buildIos/config.js`'s `ConfigConstants.swift` generator writes to the new path and adds the `canImport(UIKit)` guard — verified against real generated output, not just the checked-in stub.
- **Access-control pass**: several moved types (`URLWhitelistManager`, `VideoStreamStateMachine`/`VideoStreamState`, `NetworkMonitor`/`NetworkStatus`, `FrameworkServerUtils`) had `internal` members that broke once consumed across the new module boundary — widened symbol-by-symbol based on actual call sites, not blanket-published.
- **Test coverage**: added 61 new unit tests for the 4 files that had zero coverage anywhere in the repo (`VideoStreamState`, `VideoStreamStateMachine`, `NetworkMonitor`, `CatalystConstants`), plus `NotificationHandlerProvider` and `AnyCodable`. `CoreLogic` package: **154 tests, 0 failures**, non-`FrameworkServerUtils` line coverage ~88% (up from ~38%). Package-wide is ~55% because `FrameworkServerUtils.swift` (61% of the package's lines) is mostly real socket/`NWConnection` handling — Tier 2 (#416) scope, not chased here.

## Verified

- `swift test --package-path CoreLogic`: 154/154 pass, ~0.5s, **no simulator**
- `xcodebuild build` / `xcodebuild test -only-testing:iosnativeWebViewTests` (existing simulator path, same as CI's `test-ios-unit`): still green, no regressions
- Both re-verified from a fully clean state (`xcodebuild clean` + removing all `.build`/`.swiftpm`/`.package-config-hash`)

## Not yet done

- CI job `test-ios-logic-unit` (plan step 10) — this PR proves the split works locally; the CI job to actually get the fast-path benefit in real CI is next
- Notifications-enabled build path (`Sources/CatalystNotifications/`) is import-fixed but not locally compile-verified (needs Firebase package resolution)
- `FrameworkServerUtils.swift` loopback HTTP tests (in progress, follow-up commit)

Full plan and findings: `/Users/mayankmahavar/.claude/plans/polymorphic-fluttering-trinket.md`


mayankmahavar1mg:
Closing — this branch's content has been merged into #418 (feature/340-error-contract-tests), along with fixes made along the way (vitest min-release-age gate pin propagated through the stack). #418 stays open as the collapsed integration branch; all CI is green there.

PR #442, opened 2026-08-26
feat: Catalyst Companion app and native preview plugin

mayansharma:
A native shell for previewing Catalyst apps on a real device — scan the QR from a dev server and the app runs with the full native bridge, using the config your server provides.

Two parts: a `companion` plugin in `catalyst-core` that provides the preview runtime, and the example app in `examples/catalyst-companion` that uses it.

## The preview plugin (`catalyst-core`)

A new internal plugin under `src/native/internal-plugins/companion/`, with implementations for both platforms:

- Loads an arbitrary HTTPS app into an isolated WebView, with a URL policy governing what a previewed app may reach
- Confirmation sheet before opening a preview, a loading state while the target resolves, and shake-to-exit while inside one
- Runtime config so a previewed app gets a sane environment rather than the host app's

Supporting changes: `CustomWebview.kt`, `MainActivity.kt`, `NativeBridge.kt` on Android; `WebViewModel.swift`, `RuntimeConfig.swift`, `BridgeCommandHandler.swift` on iOS.

Also `CATALYST_QR` — an env toggle that forces the dev-server QR code on or off, so it can be scripted rather than only appearing on a TTY.

## The app (`examples/catalyst-companion`)

- Try Your Own App: full-screen QR scanner, manual URL entry, and an exit sheet
- Native capability showcase — video streaming, file picker, Google sign-in, notifications
- Bottom navigation with native transitions, and a first-launch onboarding video
- Docs links point at the deployed site rather than bundling a copy

## Notes

`fix(ios): wire CFBundleDisplayName to APP_DISPLAY_NAME` is a small unrelated fix that rode along — the iOS app name was not being read from config.

`assets.js` tightens the AndroidManifest meta-data regex from `[^>]` to `[^<>]`. The looser form could match across tag boundaries and corrupt the manifest on a second build.

No CI job for the example. An earlier revision added one; it was the wrong call for an app this repo does not ship, and the smoke script it used has been removed.

## Verification

- Companion builds clean from a fresh `npm install` + `sync-core`, lint passes
- Core E2E 6/6, including the node-loader unit tests this branch adds
- Native builds were exercised on an Android emulator and an iOS simulator during development; this PR does not add a device build to CI


mayankmahavar1mg (changes requested):
we need to see how to keep companion app related changes in native

Issue #26, opened 2025-01-09
Server crashes with unclear error when config.json is modified


mayankmahavar1mg:
## Description
When values in `config.json` are modified, the server crashes without providing a helpful error message that would indicate the cause of the failure.

### Current Behavior
- Server crashes when `config.json` values are changed
- Unclear/unhelpful error message is displayed

### Steps to Reproduce
1. Open `config.json`
2. Modify any value in the configuration and save
3. Observe the crash and unclear error message

### Screenshot
<img width="1490" alt="Screenshot 2025-01-09 at 2 57 23 PM" src="https://github.com/user-attachments/assets/95066c55-36e3-4713-aa77-ec40b167bd52" />

### Additional Context
Please provide:
- The specific error message currently shown
- Example of the config change that triggered the crash
- Server logs if available

PR #445, opened 2026-08-26
fix(core): api freeze phase 1 — crashes, dead code, alias precedence

### LINKED ISSUE #341: Story: API freeze at RC cut
## Goal

Freeze the public API surface at the 1.0.0-rc.1 cut, per RFC #177's definition of 1.0 as "ready for API commitment." Depends on Story 1's runtime-adapter decision.

## Tasks

- [ ] Lock the public API exports map in package.json
- [ ] Remove deprecated and legacy code paths
- [ ] Finalize public API and type names
- [ ] Ship TypeScript definitions for the public API
- [ ] Route any RC-discovered breaking changes through the RFC process

## Parent

Epic: #329

mayansharma:
Part 1 of 4 of the API freeze stack (#341). Review-only stack: this PR and the two after it will not be merged individually — after all four are approved, a single release PR (phase 4 branch → main, carrying the whole stack) ships as 1.0.0.

Non-breaking bug fixes:

- useNotification and useNotificationPermission: hooks hoisted above SSR guards (rules-of-hooks); SSR stub returns all 23 keys of the real hook with silent no-ops; stubs report isNative false / isWeb true and mirror useBaseHook initial state
- useVideoStream: all four return branches share one key set
- TerminalProgress.js casing rename (imports failed on case-sensitive filesystems)
- Dead code deleted: standardizedHooks.js, the router.js shim re-exporting an undeclared package, the unused scripts validator (~820 lines, zero importers verified)
- Module alias precedence unified: the app's aliases win at all three merge sites via one shared getModuleAliases(); previously vite.config let catalyst win while the SSR loader and cache-key plugin let the app win, so one import could resolve differently within a single build
- Documented the config.json → process.env loading contract (JSON types survive; shell vars are not carried over)

Review order: this PR first, then phase 2 → 3 → 4.


mayankmahavar1mg:
## CodeRabbit review — API-freeze stack, all phases against `epic/329`

Ran `coderabbit review --base epic/329` from the tip of the stack (`feat/api-freeze-p4-typescript`), so this is one cumulative pass over Phases 1–4. **61 findings total (6 critical, 35 major, 20 minor).** Each finding was checked against `epic/329` to tell *pre-existing* code (carried over by the `.js`→`.ts` renames, CodeRabbit re-scans whole renamed files) from code *this stack introduced*. Findings are posted per phase; this comment covers **Phase 1 (#445)** plus the overall picture.

### Overall stack rating: **4/5 — merge-ready after the criticals are triaged**

| Phase | PR | New-code risk | Findings attributed | Rating |
|---|---|---|---|---|
| 1 — bugfixes, dead-code, alias precedence | #445 | Low | 2 (1 major on net-new code, 1 rename-only) | 4.5/5 |
| 2 — frozen export surface, RR peer dep | #446 | Low | ~7 (mostly doc/example `react-router/dom` import nits) | 4.5/5 |
| 3 — native hook contract, app contract | #447 | Medium | hook-contract additions verified sound; flagged issues in those files are pre-existing | 4/5 |
| 4 — TypeScript for frozen surface | #448 | Low (mechanical) | ~48 — **almost all pre-existing**, surfaced by whole-file re-scan on rename | 4/5 |

**Key takeaway:** the 6 criticals and the bulk of the majors are **pre-existing bugs in `epic/329`**, not regressions from the freeze. They're real and worth fixing, but they don't block the freeze — recommend filing a follow-up hardening issue and cherry-picking any that are cheap. The stack's own new code is clean.

---

### Phase 1 (#445) findings

| # | Sev | File | Finding | Status | Verdict |
|---|-----|------|---------|--------|---------|
| 26 | Major | `src/vite/moduleAliases.js` | `getModuleAliases(appRoot)` calls `path.resolve(appRoot, "package.json")`; if `appRoot` is `undefined` this throws `TypeError`. Add a `process.cwd()` fallback so a call site with no app root still yields catalyst-only aliases. | **New in #445** (file is net-new here) | CONFIRMED — valid. All 3 current call sites pass a root, so it's latent, but the fallback is cheap and matches the pattern the type work expects. Recommend fixing in #445. |
| 8 / 48 / 49 | Major / Minor / Minor | `src/native/TerminalProgress.js` | `render()` skips a write when `output === lastRender` even after `clearLines()` erased the block; `titlePaddingTop` emits an extra newline via `console.log`; `pause()` doesn't flush `render()` before setting `isPaused`. | **Pre-existing** — #445 only renames `terminalProgress.js → TerminalProgress.js` (`0 insertions, 0 deletions`). | CONFIRMED as real behavior, but not introduced by this PR. File as follow-up or fix opportunistically. |

**Phase 1 verdict: 4.5/5.** The rename + dead-code deletion + alias unification are clean. One cheap guard to add (#26).

mayankmahavar1mg:
## API-freeze chain vs. the new `epic/329` (Story #340 + #343 landed today)

`epic/329` just absorbed the whole #340/#343 body (PRs #418, #452 — contract tests, error system, error-catalog, split CI workflow, native lint/coverage). That code was written *after* p2 froze the export surface / p4 converted `src/` to TS, but nothing had ever run the two together. I've integrated `epic/329` into the chain tip and resolved everything on branch **`integ/apifreeze-epic`** (based on `feat/api-freeze-p4-typescript` + `epic/329`).

**Local gate is green**: `catalyst-core` vitest 223 pass · `test:types` 0 · `lint` 0 · `npm run prepare` (build-dist + `tsc -p tsconfig.declarations.json` under `typescript ^7.0.2`) 0 · `create-catalyst-app` 25 · `mcp_v2` 31 · root lint 0.

### Merge-commit conflicts (5) — `7cae2a18`

| File | Resolution |
|---|---|
| `packages/catalyst-core/package.json` | Union of devDeps. Kept epic's `typescript ^7.0.2` + `vitest 4.1.9` + `@testing-library/*`; kept p4's `@types/react` / `@types/react-dom`. `scripts` + `exports` auto-merged clean (p4's `build-dist.mjs` / `strip-ts-sources.mjs` prepare, epic's `test` / `test:types`). |
| `src/scripts/loadEnvironmentVariables.js` | Took p4 — p1's call-time `appConfigPath` resolution (so `preServerInit.js` can report a missing `src_path` first). |
| `src/server/utils/validator.ts` | **Took epic's body**, re-applied p4's `: any` param annotations. Hard constraint: epic's `scripts/preflight.js:59` does `const err = validateConfigFile(...)` — consumes the validators as pure `CatalystError \| null`. p4's `try/throw/handleError` version returns `true\|undefined` and breaks the doctor command + `preflight.test.ts`. p4 contributed only types + a `REQUIRED_CONFIG_KEYS` hoist here; epic's PREFLIGHT-008 alias regex supersedes p4's `.includes("catalyst")`. |
| `src/server/renderer/handler.tsx` | Kept **p4's explicit-path imports** (`serverDataFetcher` from `web-router/…`, `getMetaData` from `web-router/utils/…`) since `index.tsx` is the frozen surface and doesn't re-export them. Added epic's genuinely-used validator symbols (`validateCustomDocument`, `handleError`). Did **not** re-add `generateCssLinkStrings` (dead import on epic and merge-base). Kept epic's `validateGetRoutes` / `validateConfigureStore` guards (F1–F9 / the `:379` reformat @trinachoudhury1mg reviewed) + p4's `: any`. |
| `package-lock.json` | Took epic's, regenerated with `npm install --package-lock-only` for p4's `@types/*` deltas (`1b03d1f5`). |

Plus a **`terminalProgress.js` casing fix** in the merge commit: p1 renamed it to `TerminalProgress.js` + updated its 2 imports; epic independently renamed base → lowercase (Linux-CI fix, `a116c891`) with 4 lowercase imports. 3-way merge picked p4's filename + epic's imports → broken on case-sensitive Linux. Renamed back to lowercase, repointed p1's 2 imports.

### Follow-up fixes — `64063cbb` (all flagged for your review)

1. **`tsconfig.json`: `jsx: "preserve"` → `"react-jsx"`.** `preserve` is fine for the type-only `tsc` pass, but vitest/esbuild + Vite's import-analysis read this config and refuse to transform JSX under `preserve` (*"invalid JS syntax"* on every `.jsx`/`.tsx` test). p4 never ran vitest against the converted tree. `react-jsx` is correct for both; Babel's `preset-typescript` ignores the option.
2. **`tsconfig.test.json`**: dropped the `"lib": ["es2022"]` override (inherit `es2022,dom,dom.iterable` — the #340 tests transitively typecheck now-`.tsx` sources using `window`/`document`); added `src/globals.d.ts` to `include` so `declare module "@catalyst/template/*"` resolves.
3. **`test/server/render.test.ts`**: `baseArgs` element-array fixtures were `string[]`; p4 typed `renderStart`'s params `ReactElement[]` (real caller `handler.tsx:241` passes `generateScriptElements()` output). `renderStart` is a pure pass-through so assertions hold at runtime — cast the fixtures.
4. **`test/server/fixtures/template/src/js/containers/App/index.jsx`**: imported `Outlet` from `"catalyst-core"` → `undefined` at render (*"Element type is invalid"*), because the frozen barrel doesn't re-export react-router. Repointed at `"react-router"`, matching what p2 already did to `templates/*/App/index.js`. **This is the 2nd epic consumer to break on the frozen barrel** (1st: `handler`'s barrel imports, resolved in the merge commit) — exactly what the "8 exports" freeze was meant to surface.
5. **`src/web-router/hooks.test.jsx`: DELETED.** modify/delete — p2 (`f7dd29c6`) deleted `src/web-router/hooks.jsx` calling `useNavigateWithTransition` a *"dead … stub"* (`() => useNavigate()`, real impl commented out). epic's #340 chain added this test to pin the stub. Merge kept the test (epic-only file) but honored p2's delete → orphaned test. Followed p2's freeze intent. **Please confirm the hook stays dropped.**

### Landing
`integ/apifreeze-epic` is pushed. Suggested: collapse #448 → #447 → #446 into `feat/api-freeze-p1-bugfixes` first, then merge `integ/apifreeze-epic` into it (brings only `epic/329` + these resolutions), and #445's CI is the first real pipeline run on the whole body against the split workflow.

_(Integration + resolution by Claude Code, session linked in the commits.)_

Issue #147, opened 2025-11-27
Request: Configurable Caching Strategy for Universal App


AshrefRaj:
Currently, apps built with the **Universal App** framework do not support programmatic cache clearing or any exposed settings to configure cache duration, invalidation logic, or freshness/staleness behavior. This creates issues for applications that are updated frequently: when a new release is deployed, outdated assets continue to be served, even when those assets are properly hashed.

I learned from a catalyst team member that the framework applies a fixed **48-hour caching strategy**, based on the assumption that webpage content does not change frequently. However, this assumption doesn’t hold true for many use cases.

I would like to request a way to **customize or override the caching behavior** to better align with different business requirements, for example, supporting shorter cache durations or allowing manual cache invalidation when new content is deployed.

AshrefRaj:
Bumping this request. 
Is there any development done in this regard?
Do we have any function that will programmatically clear the cached data at the developer’s discretion?

mayankmahavar1mg:
Closing this issue as the fix has been merged in https://github.com/tata1mg/catalyst-core/pull/197.

Appreciate @AshrefRaj  for reporting the issue.

PR #448, opened 2026-08-26
feat(core): api freeze phase 4 — typescript for the frozen surface

mayansharma:
Part 4 of 4 of the API freeze stack (#341), stacked on the phase 3 PR. Shows only phase 4's five commits. After approval of the full stack, this branch becomes the single release PR into main, shipped as 1.0.0.

- Consumer-facing contract migrated to TypeScript (index, web-router, router, server, native bridge — 68 files, all git renames); vite/, scripts/, the native CLI and observability files stay JS by decision, deferred post-1.0
- Build pipeline compiles instead of raw-copying dist; output proven byte-identical at the swap, and the native babel pass gains --extensions (babel silently skips .ts by default)
- Shipped .d.ts for the frozen surface with types conditions and node10 typesVersions; the 16 deprecated alias keys carry @deprecated markers visible in consumers' editors
- scripts/check-consumer-types.mjs (npmignored): packs the tarball into a scratch consumer and runs 17 assertions across three resolution modes, including negatives pinned to specific TS error codes


mayankmahavar1mg:
## CodeRabbit review — API-freeze stack (Phase 4 of 4)

From the cumulative `coderabbit review --base epic/329` run at this branch tip. Full context and Phases 1–3 tables on #445. **49 of the stack's 61 findings land here** — because Phase 4 renames ~60 files `.js`→`.ts`/`.jsx`→`.tsx`, and CodeRabbit re-reviews every renamed file top-to-bottom.

### Phase 4 rating: **4/5 — merge-ready; findings are a hardening backlog, not regressions**

I checked every finding against `epic/329`. **Result: essentially all of them target code that already exists verbatim in `epic/329`** — the rename carried it over untouched. Phase 4's *actual* diff (type annotations, `.d.ts` generation config, a few signature tweaks) is mechanical and low-risk. The 6 criticals are **pre-existing bugs**, worth fixing but not blockers for the freeze.

**Recommendation:** land Phase 4 as-is for the freeze; open a linked "1.0 hardening" issue carrying the criticals + payload-normalization majors. Fix in-stack only the cheap type-correctness ones (#33 `errorCode`→`statusCode`, #35 implicit-any params) since Phase 4 owns typing.

### Legend
- **Pre-existing** = identical code in `epic/329`; surfaced by whole-file re-scan on rename. Not introduced by this PR.
- **New/typing** = introduced or newly relevant because of Phase 4's type work.

### Critical (6) — all pre-existing

| Finding | File | Status |
|---|---|---|
| Rules-of-Hooks: `if (base.webFallbackDisabled) return` sits above `useCallback(webTakePhoto)`; toggling `setWebFallback` changes the hook count between renders → React crash | `hooks/useCamera.ts:89` | Pre-existing (identical in epic/329) |
| Same Rules-of-Hooks pattern: early return above `useEffect` in `useHapticFeedback` | `hooks/useHapticFeedback.ts:69` | Pre-existing |
| `_showWebOverlay` appends a new overlay without removing `overlayRef.current` from a prior re-entrant navigate → orphaned fixed-position `<div>` covering the page | `hooks/useNativeTransition.ts:103-115` | Pre-existing |
| `useNotificationPermission` registers directly with `WebBridge` instead of `registerPermissionStatusListener`; shared listener key never unregistered when the set empties → leak | `hooks/useNotification.ts:111-115` | Pre-existing |
| Inline `<script>` writes `__INITIAL_STATE__` / `__ROUTER_INITIAL_DATA__` / `safeArea` without escaping `<`, `>`, U+2028/U+2029 → XSS / script-termination if serialized data contains `</script>` | `server/renderer/document/Body.tsx:38-47` | Pre-existing (XSS — prioritize in the hardening issue) |
| `RouterDataProvider` reads `config.disableCaching` when `config` prop is `undefined` → `TypeError` in `shouldFetch` | `web-router/components/RouterDataProvider.tsx:285` | Pre-existing |

### Major (32) — grouped

**Payload / input hardening (native bridge)** — all pre-existing; bundle into hardening issue:
| Finding | File |
|---|---|
| `parseDeviceInfo` in legacy `ON_DEVICE_INFO_SUCCESS` not wrapped in try/catch → promise left pending on malformed payload | `native/bridge/WebBridge.ts:380-383` |
| `handlePermissionStatus` builds rejection error from raw non-string payload → `TypeError`, promise never settles | `hooks/useCamera.ts:265-269`, `hooks/useNotification.ts:55-62` |
| `useGoogleSignIn` throws during render when `window.WebBridge` missing (other hooks return an error shape) | `hooks/useGoogleSignIn.ts:26-28` |
| `useDataProtection.webClearWebData` IndexedDB delete never resolves on `blocked` event → `Promise.all` hangs | `hooks/useDataProtection.ts:58-61` |
| `useSafeArea` passes `env(safe-area-inset-*)` as a property name to `getPropertyValue()` (should read the `--sat/--sar/--sab/--sal` custom props) | `hooks/useSafeArea.ts:43-46` |
| `CameraUtils.requestCameraPermission` has no timeout / error-callback → promise pending forever if native never responds | `native/bridge/utils/CameraUtils.ts:116-128` |
| `isValidFormat/Device/FlashMode` can return non-boolean on empty input | `native/bridge/utils/CameraUtils.ts:212` |
| `FileObjectConverter` loopback check is substring-based → `localhost.evil.com` passes; parse URL and compare exact hostname | `native/bridge/utils/FileObjectConverter.ts:79-81` |
| `NativeBridge.call` uses constructor-captured `isAndroid/isIOS`; a late-injected WebView bridge falls through to `return false` | `native/bridge/utils/NativeBridge.ts:136-142` |
| `userAgentUtil.getUserAgentDetails` passes non-string `ua` into `.includes()` helpers | `server/utils/userAgentUtil.ts:86-90` |

**Perf collectors** — all pre-existing; robustness against malformed native events:
| Finding | File |
|---|---|
| `notify*` formatters call `.toFixed()` on possibly-missing numeric fields; `url.split()` on non-string | `perf/collectors/InsightsCollector.ts:241` |
| `RenderCollector.onNativeEvent` — guard `minFps` / `jvmMb` / `webviewMb` / `totalMb` before formatting | `perf/collectors/RenderCollector.ts:88` |
| `InteractionCollector` start mark isn't per-interaction unique → delayed callbacks resolve the wrong mark | `perf/collectors/InteractionCollector.ts:168` |
| `NavigationCollector._setupLcpFinalization` re-registers `visibilitychange` on every SPA route change → duplicate listeners | `perf/collectors/NavigationCollector.ts:90-92` |
| trace-export counter event name varies per memory sample (should be a stable constant; label goes in `args`) | `perf/core/traceExport.ts:132` |

**SSR renderer (per-request global state)** — pre-existing; real concurrency hazards under load:
| Finding | File |
|---|---|
| `ChunkExtractor` registered on global `__CHUNK_EXTRACTOR__` → concurrent requests share one extractor; pass via SSR context instead | `server/renderer/ChunkExtractor.ts:35-37` |
| `_renderMarkUp` uses `globalThis.__SAFE_AREA_INITIAL__` with save/restore → races across concurrent renders; pass via `SsrRequestProvider` like `isBot` | `server/renderer/handler.tsx:209-218` |
| `tracedRenderMarkUp` inside the fetcher try/catch → render failures mis-logged as `FETCHER` and trigger the 404 fallback | `server/renderer/handler.tsx:437-447` |
| post-shell `onError` after `onShellReady` destroys `tail` + rejects but doesn't end the HTTP response → hung connection | `server/renderer/handler.tsx:349-350` |
| `manifestCache` builds `buildPath` from `process.env.src_path` at module load, outside the try → crash on missing config | `server/manifestCache.ts:15` |
| `expressServer` `uncaughtException` handler logs `process.stderr.fd` and doesn't exit → supervisor can't restart | `server/expressServer.ts:99-102` |
| `expressServer` `app.listen` callback takes an `error` arg that's never populated; real bind errors (`EADDRINUSE`) go unhandled — attach `server.on('error')` | `server/expressServer.ts:285-292` |
| `render.js/ts` `renderEnd` returns `errorCode`, handler expects `statusCode` → status never propagates to `Body` | `server/renderer/render.ts:44` |
| `renderStart` `JSON.parse(IS_DEV_COMMAND)` on unset env throws; `PUBLIC_STATIC_ASSET_URL` + `_PATH` unset → `"undefinedundefined"` asset path | `server/renderer/render.ts:16-20` |

**web-router:**
| Finding | File | Status |
|---|---|---|
| `MetaTag` effect doesn't reset state when `mergedMetaTags` is empty → previous route's tags persist | `web-router/components/MetaTag.tsx:19-21` | Pre-existing |
| `RouterDataProvider` effect deps use fresh `match.params` object → effect fires every render | `web-router/components/RouterDataProvider.tsx:292` | Pre-existing |
| `delete currentPageData.fetcherNotAvailable` mutates the shared `RouterContext` object | `web-router/components/RouterDataProvider.tsx:345` | Pre-existing |
| `Split` spreads `{...props}` after internal `onVisible/skipVisibility/isBot` → consumer can override framework internals | `web-router/components/Split.tsx:212-218` | Pre-existing |

**Typing (new / relevant to Phase 4 — fix in this PR):**
| Finding | File |
|---|---|
| `render.ts` `renderStart`/`renderEnd` destructured params implicitly `any` — add explicit types | `server/renderer/render.ts:4-13, 35` |

### Minor (11)
`useDeviceInfo` screen units mismatch web-fallback vs `WebBridge.getDeviceInfo` (`devicePixelRatio` double-applied); `useFilePicker.options.multiple` not coerced to boolean; `useHapticFeedback` success not normalized to boolean (#147); `useVideoStream` webFlip loses fps constraint; `HookCollector` lazy-span can stay open indefinitely under sustained LoAF; `summary.ts buildSummary` helper failures on partial collections; `traceExport` Android branch silent return; `useBaseHook.startProgress` uses `"active"` instead of `PROGRESS_STATES.STARTING`; `expressServer` `OTEL_ENABLE` doesn't accept string `"true"`; `cjsRequire` doc example wrong; `metaDataUtils.mergeHeadElements` key collision between `charSet`/`name`/`property`/`title`. **All pre-existing except the doc/example nits.**

---

**Phase 4 verdict: 4/5.** The rename+typing diff itself is clean and merge-ready. The 49 findings are a real but pre-existing hardening backlog — recommend a linked "Catalyst-Core 1.0 hardening" issue for the criticals and the SSR-global-state + payload-normalization majors, and fixing only the type-correctness items (`errorCode`→`statusCode`, implicit-any params) in this PR.

PR #472, opened 2026-09-11
feat(otel): add lean memory heartbeat and trace buffer diagnostics

vishalpolley-1mg:
- Add periodic 60s memory and span processor heartbeat (`[OTEL Heartbeat]`) for Last9 tracking
- Add bounded cache & overflow warning (`[OTEL Buffer Overflow]`) when `promotedTraces` exceeds 1024 entries
- Add throttled high-memory warning (`[OTEL High Memory Warning]`) when RSS > 800MB or HeapUsed > 500MB
- Add memory snapshot on process exit (`[OTEL Shutdown]`) during SIGINT/SIGTERM
- Add safe logger wrapper (`getLogger()`) with fallback for loggers missing `.warn()`
- Support boolean and string formats for `process.env.OTEL_ENABLE`

### DIFF
--- a/src/otel.js
+++ b/src/otel.js
@@ -28,6 +28,41 @@ export const IS_BOT_KEY = createContextKey("catalyst.is_bot")
 import semanticConventions from "@opentelemetry/semantic-conventions"
 const { ATTR_SERVICE_NAME, ATTR_SERVICE_VERSION, ATTR_DEPLOYMENT_ENVIRONMENT } = semanticConventions
 
+const formatMB = (bytes) => (bytes / 1024 / 1024).toFixed(1)
+
+function getLogger() {
+    const l =
+        (typeof logger !== "undefined" && logger) ||
+        (typeof global !== "undefined" && global.logger) ||
+        console
+    return {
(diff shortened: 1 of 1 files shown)


src/otel.js
```diff
@@ -22,12 +22,90 @@ import { OTLPMetricExporter } from "@opentelemetry/exporter-metrics-otlp-grpc"
...
+    }
+}
+
+// K8s pods have a per-container memory limit that os.totalmem() can't see (it
+// always reports the node's full memory) — read the cgroup limit directly
+// (v2, then v1), falling back to host memory outside a container. Cached: the
+// limit can't change during the process's lifetime, and both thresholds below need it.
+let memoryLimitBytes
+function getMemoryLimitBytes() {
+    return (memoryLimitBytes ??=
```
trinachoudhury1mg, line 64: 1. Memory Request instead of limits - set threshold offset based on memory autoscaling for the pod
2. Memory.current - should also be logged to give insight to ratio between memory.current / memory.request 
3. enableMemoryDiagnostics only makes sense to have incase of sampling

trinachoudhury1mg (changes requested):

trinachoudhury1mg (approved):

trinachoudhury1mg (approved):

